PR #19, opened 2024-11-29
fix integration test and further break down casting

AVSuni:
The PR fixes broken integration test and further breaks down casting. Now we can add test cases and chainsaw tests to validate inputs and outputs in the integration test.


punasusi (approved):
Nice! Tested a clean install, and some CRD setup timing is only thing (but not new).

PR #37, opened 2024-12-15
packaging

punasusi:
Move to Images for packageing and deployment


woojae-siloai:
Is the order of commands "smelt", "cast", "docker build", "forge"?
Grafana was deployed when I followed the above order?

punasusi:
There is no docker-build step, but it's part of the cast step. The reason we ran in separately was to test where the error was coming for the certificate issues we saw.

templates/stack.yaml
```diff
@@ -5,3 +5,4 @@ metadata:
   namespace: crossplane-system
 spec:
   domain: demo.silogen.ai
+  storageClass: directpv-min-io
```
woojae-siloai, line 8: Will "directpv" come from a k8s cluster itself before crossplane does its jobs?

woojae-siloai (approved):

PR #54, opened 2025-01-13
added grafana dashboard for GPU's

punasusi:


brownzebra (approved):
tested that it does indeed deploy

PR #60, opened 2025-01-22
Shift to use ArgoCD instead of Crossplane as internal controller

punasusi:


brownzebra (approved):
Tested. I'll create issue to update README. Also should look at known issues in there.

PR #64, opened 2025-01-26
updated docs and devbox

punasusi:


docs/DEVBOX.md
```diff
@@ -0,0 +1,72 @@
+<!-- gen-readme start - generated by https://github.com/jetify-com/devbox/ -->
+## Getting Started
+This project uses [devbox](https://github.com/jetify-com/devbox) to manage its development environment.
+
+Install devbox:
+```sh
+curl -fsSL https://get.jetpack.io/devbox | bash
+```
+
+Start the devbox shell:
```
woojae-siloai, line 10: What about to add like this?
Start the devbox shell at root folder of cluster-forge having devbox.json

punasusi: this devbox.md was auto-generated, so I would prefer to keep it as is, but maybe the /docs is not the right place...

PR #66, opened 2025-01-27
needed files for ansible to setup cluster

punasusi:


brownzebra (approved):
ROCm tested on OCI2 and seems ok

PR #82, opened 2025-02-05
Change sourcefile to manifestpath and allow directories

mramdgh:


punasusi (approved):
needs testing, but looks great!

punasusi (approved):

PR #85, opened 2025-02-07
163 shift fully to longhorn

mramdgh:


woojae-siloai:
Just checked longhorn with manifest and manual by running kubectl command. I think it will work with forge command too.

woojae-siloai (dismissed):

brownzebra:
seems legit

punasusi (approved):
YES!!

PR #103, opened 2025-02-18
minio use of longhorn and lgtm monitoring

punasusi:


input/minio-tenant/values.yaml
```diff
@@ -29,7 +29,7 @@ tenant:
       # If using Amazon Elastic Block Store (EBS) CSI driver
       # Please make sure to set xfs for "csi.storage.k8s.io/fstype" parameter under StorageClass.parameters.
       # Docs: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/parameters.md
-      storageClassName: longhorn-default
+      storageClassName: longhorn-minio
```
woojae-siloai, line 32: Does longhorn have this storageclass?

punasusi: Yes. Not the one in ClusterForge, but the one in the install script we are moving to does have this.

woojae-siloai (approved):

PR #105, opened 2025-02-19
Update document external secrets

woojae-siloai:


punasusi:
At first glance, this looks GREAT and correct. I think let's still have a team-wide sync on how we will ongoing populate external secrets in all the deployment target environments we are aiming toward.

punasusi (approved):

PR #106, opened 2025-02-19
Add collection in config.yaml

woojae-siloai:
I guess this will help this [ticket](https://github.com/silogen/platform/issues/142) to bundle related tools.

```
Specify three sets of cluster tools for monitoring, so that cluster forge users can easily setup the monitoring they need

minimal, collecting logs, showing gps are active, not storing anything
external 1 + push to an external service, whatever customer X happens to have
third 1 + push to our internal ops cluster
```
@punasusi just want to get a quick review before open PR


punasusi:
Yes, this looks good! Then we can add collections and still manage the individual components as separate objects. The theory here looks correct, and would fulfill the desired uses :) Great!

woojae-siloai:
@punasusi  FYI

I ran "smelt" and selected "all" at both of "main" and "add-collection-config-yaml" branches with/without devbox. I ran "just clean-all" every time.

I found that only selected "all" with devbox gave all the tools. I cannot get all the tools without devbox.

PR #123, opened 2025-02-20
less gum

punasusi:


brownzebra (approved):
I find the gum working pleasingly

PR #126, opened 2025-02-21
Consolidate scripts and the yolo script host location

punasusi:


brownzebra (approved):
Nice! Will test today.

PR #130, opened 2025-02-25
Add sidecar to add grafana dashboards to LGTM stack automatically

woojae-siloai:
This PR adds a sidecar container to lgtm pod to load grafana dashboards from configmaps dynamically and split one file into three manifests files. It is tested on a kind cluster with "standard" storageclass. Grafana dashboards are added as expected. Dashboards under /otel-lgtm-stack/dashboards/ are going to be added automatically. It is possible to add dashboard from a configmap dynamically by kubectl apply command if the configmap has proper label and annotation like in  /otel-lgtm-stack/dashboards folder.

Next things to do:
The names of labels and metrics which are expected to come to prometheus and are shown by the grafana are different from those collected by [otel-collecter-agent](https://github.com/silogen/cluster-forge/blob/69773c78007bb1c78cfab52be5072664e2f01a49/input/otel-lgtm-stack/otel-collector-agent/collector-manifests.yaml). So added dashboards can be seen, but they don't show their graphs.

Following PR will be for updating otel-collector-agents.


punasusi (approved):
Do I dare comment LGTM on an LGTM stack PR? :)

PR #133, opened 2025-02-25
Adding missing metrics configmap

rodrodsilo:
Now with the provided configmap the GPU operator should should produce some extra metrics that are not set with default metrics component


punasusi (approved):
Nice!

PR #145, opened 2025-03-03
Add option to build private image

mramdgh:


woojae-siloai (approved):
Checked out to private-option and ran `cast --private` with devbox. It created a private image using ttl.sh

PR #154, opened 2025-03-05
Update deviceconfig_example.yaml

rodrodsilo:
This will fix the deviceconfig with the correct attribute name


brownzebra (approved):
Nice!

PR #157, opened 2025-03-06
Update workflow to add ingress to release

woojae-siloai:
This PR updates workflow to add ingress to release and add wait_for_namespaces function to setup-forge.sh script before applying ingress.yaml line. I guess that 300 seconds for checking the existence of each namespace will be enough based on my experience observing the argocd app so far.

The order will be...
- run `sudo sed -i "s/{{CONTROL_IP}}/$MAIN_IP/g" ingress/ingress.yaml`  inside if clause to use `$MAIN_IP`
- run `wait_for_namespaces function`
- run `KUBECONFIG==$HOME/.kube/config kubectl apply -f ingress/ingress.yaml` outside of if clause


punasusi (commented):
looks good!! I'll leave it for @brownzebra to validate on an upcoming test round to approve

brownzebra (approved):

PR #167, opened 2025-03-12
add resources.requests and resources.limits for lgtm stack

woojae-siloai:
This PR adds requests and limits of resources to LGTM stack.
LGTM-stack was blocked on a cluster having kyverno policy.

It could be good that numbers of requests and limits updated later on based on actual usages.


punasusi (commented):
We don't need CPU limits, request is enough, for memory, both are needed.

mramdgh (approved):
LGTM

PR #172, opened 2025-03-14
Delete a manifest creating external-secrets namespace

woojae-siloai:
This PR deletes a manifest in 1password-cluster-secret-store folder, which creates a "external-secret" namespace.
This gets rid of "out of sync" from the both "external-secret" and "1password-cluster-secret" tools trying to handle "external-secret" namespace.

From now on, "external-secret"(syncwave: -1) tool should be deployed together with "1password-cluster-secret-store"(syncwave: 0). Otherwise, it will fail to deploy "input/1password-cluster-secret-store/manifests/Service_central-onepass-connect.yaml"


mramdgh (approved):
1password does imply external secrets so this is excellent

PR #173, opened 2025-03-14
add otel-lgtm-stack namespace under exclude of kyverno policy

woojae-siloai:


punasusi (approved):
this is sort of ok and approved.

The better plan is to update these resources so they follow community best practices and have resources defined. This does mean more work for us to maintain, but now our team is big enough, it might be the right time. 
Do this way for now, and open a ticket for us to plan how to manage.

PR #184, opened 2025-03-20
Temporary proxy configuration

brownzebra:
Temporary proxy update


setup-cluster/setup-forge.sh
```diff
@@ -280,6 +282,25 @@ generate_longhorn_disk_string() {
     fi
 }
 
+update_proxy() {
```
punasusi, line 285: /usr/local/lib/systemd/system/rke2-server.env

This is the file where the proxy needs to be set.

punasusi (approved):

PR #193, opened 2025-03-26
first kaiwo inputs

salexo:
Adding Kaiwo to the available input options. This time manually, but working towards automating this


salexo:
Opening up PR now to check my understanding of what the goal was

AVSuni (changes requested):
Nice work! Manifests themselves look correct. I can see our patches in the two kustomized dependencies (Kueue and AppWrapper) and Kaiwo manifests also look complete with static manifests appended to the end (also correct image tag).  Just two small changes requested:

1. Replace Kueue with out patched version. Even if people don't want to use Kaiwo, the patched version enables things like pod management which is necessary for Kueue to manage Deployments (also added in the patched version)
2. also move AppWrapper operator to its own directory (i.e. let's give up kaiwo-dependencies directory)
3. Create a kaiwo "collection" that includes cert-manager, kuberay. kueue, appwrapper and kaiwo. There are examples at the end of `input/config.yaml` such as `minio-all-together`

AVSuni (approved):
Nice work! Tested and works 💪

PR #207, opened 2025-04-02
make minio-tanant scrapable

woojae-siloai:
This PR makes otel-collectors-metrics, which is a part of otel-lgtm-stack, scrape metrics from minio-tenant.
Setting `MINIO_PROMETHEUS_AUTH_TYPE="public"` is to omit passing bearer-token in scrape job block.
This change was tested on a kind cluster. 

Expected:
Minio metrics were scraped and shown from grafana.

Not expected:
Applied ready-made some minio dashboards too. Some minio metrics(for example, disk_storage_total, minio_node_drive_total_bytes) had no time series. So some panels relying on the metrics having no time series didn't show graphs.

Reason: 
Due to the kind cluster using virtual in virtual?

After approval and merging this pr, updating minio-tenant and otel-collector-metrics for oci-clusters and testing ready-made dashboard will follow next


punasusi (approved):
🚀

PR #211, opened 2025-04-07
update kaiwo to v0.1.2

AVSuni:


input/kaiwo/install.yaml
```diff
@@ -65049,6 +65049,9 @@ spec:
           requests:
             cpu: 100m
             memory: 128Mi
+          limits:
+            cpu: 2
```
punasusi, line 65053: I would remove the CPU limit, it's not required, and can't really help. But, likely little harm if we want to get this one out and address it later.

punasusi (approved):

input/kaiwo/install.yaml
AVSuni: Ok thanks. I thought it was now required.

PR #215, opened 2025-04-16
expose prometheus as a service

salexo:
Prometheus is included in the LGTM stack image, but is not exposed. This PR adds the Prometheus port to the LGTM deployment and service.


woojae-siloai (approved):
LGTM

PR #222, opened 2025-04-25
Reduce stack directories + remove setup-forge.sh

oskarasbrink:


brownzebra (approved):
tested at package level

PR #226, opened 2025-04-30
stop using release type as name for folder

oskarasbrink:
renaming to 'clusterforge'. Makes it way easier for clusterbloom


punasusi (commented):
If we are hardcoding the PACKAGE_DIR, to a fixed string, does it even need to be a variable then?

oskarasbrink:
> If we are hardcoding the PACKAGE_DIR, to a fixed string, does it even need to be a variable then?

good point, fixed

punasusi (approved):

PR #228, opened 2025-05-06
Add external-secrets calling secrets from 1password

woojae-siloai:
This PR add three external-secrets for fetching secrets from 1password for monitoring tools in ociops.

After this PR, follow-up job will be updating monitoring tool side external-secrets to fetch from the secrets in cf-es-backend, not from "fake-cluster-secret-store"


mramdgh (approved):
LGTM

PR #236, opened 2025-05-13
updated docs

punasusi:


mramdgh (changes requested):
I noticed a bunch of old leftovers, I'll need to do a more careful read-through, I'll take this over.

mramdgh (approved):
LGTM

PR #241, opened 2025-05-15
Small fixes: change the name of the ClusterRoleBinding and the UID for the Grafana dashboard

woojae-siloai:
This PR fixes two things down below:
- Changing the name of the clusterrolebinding to avoid from an error message from argocd apps ("ClusterRoleBinding/grafana-clusterrolebinding is part of applications argocd/grafana and otel-lgtm-stack")
- Changing the name of the UID because there is now only one prometheus datasource (mimir)


mramdgh (approved):
LGTM

PR #253, opened 2025-05-27
rename tw release to weka

oskarasbrink:


brownzebra (approved):
secret still needs added

PR #259, opened 2025-05-28
Update config.yaml for prometheus-app-config

oskarasbrink:
need to also update input/config.yaml for yesterdays new input prometheus-app-config. 

This look ok?


brownzebra (approved):
testing in eai

PR #304, opened 2025-07-19
Fix quay.io image SHA resolution for multi-architecture images

punasusi:
- Updated getQuaySHA() to handle manifest lists properly
- Added support for platform-specific manifest extraction
- Fixed issue where manifest list digest was returned instead of pullable platform digest
- Maintains backward compatibility with single manifests
- Resolves "image not found" errors when pulling quay.io images with SHA references


brownzebra (approved):
to testing

PR #379, opened 2025-08-26
feat: one release with all assets and semantic versioning ci on PR

brownzebra:


mramdgh (approved):
LGTM

.github/workflows/release.yaml
```diff
@@ -1,114 +1,150 @@
...
-  contents: write
-  packages: write
+      - name: Get latest version
+        id: get_version
+        run: |
+          # Get latest version tag or default to v0.0.0 if none exists
+          CURRENT_VERSION=$(git describe --tags --match "v[0-9]*" --abbrev=0 2>/dev/null || echo "v0.0.0")
+          echo "current_version=${CURRENT_VERSION}" >> $GITHUB_OUTPUT
+
+      - name: Calculate new version
```
Q-Dub, line 42: Doesn't need to be for this initial iteration, just flagging that we will want to something more robust as opposed to reinventing the wheel for some of the logic, namely: https://github.com/semantic-release/semantic-release (built-in support for conventional commits).

PR #384, opened 2025-08-27
SDA-2177 Cluster-Forge changes for KC ClientScope Group

Klevmarken:
Adds a Client Scope "groups" to the AIRM UI Client and sets it to default to include the group membership of its clients in the JWT token.


rodrodsilo (approved):
lgtm
It looks like you are going to enable Access Control on airm👍

brownzebra (approved):
to test release

PR #395, opened 2025-09-08
Set kueue mutating webhook reinvokation policy to IfNeeded

akshayv:
Set kueue mutating webhook reinvokation policy to IfNeeded so it can work with Kyverno https://github.com/kubernetes-sigs/kueue/issues/4681


AVSuni (approved):
🥇

PR #418, opened 2025-09-29
Remove unneeded secret

akshayv:


silomedvedev:
Not needed since: https://github.com/silogen/cluster-forge/pull/478

PR #641, opened 2026-03-18
feat: add options to disable selfHeal

tomastola:
## Summary

- Add option to disable ArgoCD `selfHeal` on cluster-apps and cluster-forge root apps — required for e2e tests where ArgoCD would otherwise revert test-created resources
## Test plan

- [ ] Verify cluster-forge deploys successfully with selfHeal enabled (default behaviour unchanged)
- [x] Verify e2e tests can set `selfHeal: false` via values to prevent ArgoCD reverting test resources
- [x] Verify airm/aiwb apps no longer show as out-of-sync due to ExternalSecret operator fields

### DIFF
--- a/root/templates/cluster-apps.yaml
+++ b/root/templates/cluster-apps.yaml
@@ -52,7 +52,7 @@ spec:
   syncPolicy:
     automated:
       prune: true
-      selfHeal: true
+      selfHeal: {{ hasKey $scope "selfHeal" | ternary $scope.selfHeal true }}
     syncOptions:
       - CreateNamespace=true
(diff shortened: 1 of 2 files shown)


punasusi (commented):
Why are we using argocd to deploy applications that we don't want argo to manage?
If we are running end-end tests, we can kubectl apply the manifests, and create new ExternalSecrets which are not managed, and then can be deleted and managed outside of the argo/gitea scope, rather than adapting the production delivery templates with test if/then code.

brownzebra:
Per comments above, closing

PR #688, opened 2026-04-14
EAI-2243 remove hardcoded password for gitea devuser

woojae-siloai:
This PR makes a devuser account for gitea to use a random password from OpenBao.
- OpenBao creates gitea devuser password first
- Init job fetches and creates devuser account
- External-secret creates the dedicated secret later


mramdgh (approved):
LGTM

PR #693, opened 2026-04-23
Eai 5838 optimizing down the small configuration to fit in

brownzebra:


brownzebra:
merged into EAI-1470-migrate-cluster-forge-deployment-logic-to-cluster-bloom

PR #696, opened 2026-04-28
EAI_2309 Openbao init fix

woojae-siloai:
Summary: Fix OpenBao Init Job Rerun on TargetRevision Changes

Problem
When updating targetRevision in cluster-values.yaml, the openbao-init-job would rerun unnecessarily, causing the openbao-config application to get stuck in a progressing state. This violated the design principle that initialization should only occur once during bootstrap.

Root Cause
The openbao-init-job was incorrectly managed as a persistent ArgoCD application in cluster-forge/root/values.yaml. This meant every targetRevision change triggered ArgoCD to sync the init job, causing it to rerun when it should only execute during initial cluster bootstrap.

Solution
Removed openbao-init from ArgoCD application management while preserving its bootstrap functionality:
- Removed openbao-init application definition from values.yaml
- Removed from enabledApps lists in size-specific configuration files
- Removed from SBOM components list
- Preserved existing bootstrap logic in scripts/bootstrap.sh

Result
- ✅ Init job runs only once during bootstrap
- ✅ No rerun on targetRevision changes
- ✅ openbao-config no longer gets stuck in progressing state

### DIFF
--- a/root/values_large.yaml
+++ b/root/values_large.yaml
@@ -45,7 +45,6 @@ enabledApps:
   - minio-tenant-config
   - openbao
   - openbao-config
-  - openbao-init
   - opentelemetry-operator
   - otel-lgtm-stack
   - prometheus-crds
--- a/root/values.yaml
+++ b/root/values.yaml
@@ -745,15 +745,7 @@ apps:
     path: openbao-config/0.1.0
     syncWave: -60
     valuesFile: values.yaml
-  openbao-init:
-    helmParameters:
-      - name: domain
-        value: "{{ .Values.global.domain }}"
-    namespace: cf-openbao
-    path: ../scripts/init-openbao-job
-    syncWave: -50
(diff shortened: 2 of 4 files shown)


root/values_large.yaml
```diff
@@ -45,7 +45,6 @@ enabledApps:
   - minio-tenant-config
   - openbao
   - openbao-config
-  - openbao-init
```
oskarasbrink, line 48: should this openbao init not be run anymore in its own app? if it fails in the first run, how would we re-run it?

oskarasbrink (approved):

root/values_large.yaml
woojae-siloai: I think openbao init should be run by bootstrap that makes this job only run once.
So re-run bootstrap.sh would be the answer.

PR #699, opened 2026-04-29
EAI-6059 Prepare Pluggable S3 instructions for changes in upstream

pre:
EAI-6059

Goal: Prepare Cluster Forge side of Helm instructions to the point where next step is to create commits to other repositories than cluster-forge. 

## Summary

  Move the live parts of `scripts/s3.sh` into `install_base.sh`'s `PLUGGABLE_S3=true` branch so the install path is declarative end-to-end and `s3.sh` shrinks  to a non-destructive post-install verifier. Same shape applied to secret handling: CNPG and MinIO Secrets split out of `secrets-aiwb.yaml` so they can be applied conditionally per pluggable mode, with env-based replacements created in `install_base.sh` when in pluggable mode.

  Components touched:
  - **Secrets**: new `secrets-aiwb-cnpg.yaml` / `secrets-aiwb-minio.yaml`; `install_base.sh` applies each only when the corresponding in-cluster component is   being installed; in pluggable mode it creates env-based replacements (`AIWB_DB_SECRET_NAME`, `KEYCLOAK_DB_SECRET_NAME`, `minio-credentials` in `aiwb` and
  `workbench`).
  - **PLUGGABLE_S3 install_base.sh branch**: redirect Service `minio.minio-tenant-default.svc.cluster.local` → `${MINIO_HOST_IP}:${MINIO_PORT}` is now created   here (previously in `s3.sh`); idempotent `kubectl delete svc/endpoints` before apply so upgrades from `PLUGGABLE_S3=false` don't hit SSA conflicts.
  - **`scripts/s3.sh`**: reduced from ~320 → ~120 lines. Tenant deletion, Secret patches, namespace label fix, and `kubectl set env` / restart all removed.   Remaining script just prints AIWB's MinIO env, the `minio-credentials` access keys, the redirect Endpoints, and runs a curl pod against `/minio/health/live`.
  - **`scripts/pluggable.sh`**: `${PLUGGABLE_*:-false}` so pre-exported values pass through (previously unconditionally set to false).
  - **Docs**: `components/db.md` and `components/s3.md` rewritten for the `pluggable.sh`-based flow; `s3_minio_container.sh` instructions point at   `scripts/pluggable.sh` instead of the broken `pluggable/scripts/s3.sh` path.

The remaining redirect Service workaround exists because `silogen/aim-performance` workloads hardcode `BUCKET_STORAGE_HOST` to the in-cluster MinIO URL. Tracked separately as a Phase 3 follow-up across `silogen/core` and `silogen/aim-performance`; once those land, the redirect block and `s3.sh` can be removed.

  ## Test plan

  - [x] `PLUGGABLE_DB=false PLUGGABLE_S3=false ./scripts/install_base.sh <DOMAIN>` — in-cluster CNPG + MinIO path, all Secrets applied, no env-based block  triggers
  - [x] `PLUGGABLE_DB=true ./scripts/pluggable.sh <DOMAIN>` with `POSTGRES_HOST` etc. set — `secrets-aiwb-cnpg.yaml` skipped, `${AIWB_DB_SECRET_NAME}` /  `${KEYCLOAK_DB_SECRET_NAME}` created from env, AIWB and Keycloak come up against external Postgres
  - [x] `PLUGGABLE_S3=true ./scripts/pluggable.sh <DOMAIN>` with `MINIO_HOST` etc. set — `secrets-aiwb-minio.yaml` skipped, `minio-credentials` created from env   in `aiwb` and `workbench`, redirect Service points at `${MINIO_HOST_IP}:${MINIO_PORT}`, AIWB talks to external MinIO

### DIFF
--- a/docs/manual_helm_install/components/db.md
+++ b/docs/manual_helm_install/components/db.md
@@ -1,272 +1,110 @@
-# Pluggable PostgreSQL Compatible Database
+# Pluggable PostgreSQL Database
 
-This guide describes how to replace the ClusterForge-managed PostgreSQL database with your own external PostgreSQL instance.
-
-## Overview
-
-ClusterForge provisions your Kubernetes cluster with AIWB and Keycloak already deployed and connected to an in-cluster PostgreSQL database managed by CloudNativePG (CNPG). The AIWB and Keycloak deployments are pre-configured with hardcoded database usernames that match this CNPG installation.
-
-These instructions walk you through:
-
-1. Setting up your own external PostgreSQL server with the required databases and users
-2. Removing the CNPG cluster resources from the `aiwb` and `keycloak` namespaces
-3. Patching the Kubernetes secrets and deployments that ClusterForge created so that AIWB and Keycloak connect to your database instead
+Cluster Forge can be installed against an external PostgreSQL server instead
+of the in-cluster CloudNativePG (CNPG) Clusters. In this mode `install_base.sh`
+skips the cnpg-operator install, skips the Cluster CRs in the AIWB and
+Keycloak charts, and creates the user Secrets that AIWB and Keycloak read at
+startup directly from environment variables.
 
 ## Prerequisites
 
-1. **PostgreSQL 12 or later** is installed and running
-2. **kubectl** is installed and configured to access your ClusterForge cluster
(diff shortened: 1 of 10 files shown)


Q-Dub (approved):
LGTM, thanks! 🚀

PR #702, opened 2026-04-30
EAI_5630_fix "--aiwb-only" flag

woojae-siloai:
Summary

Problem Fixed
AIWB standalone deployments failing due to missing domain configuration in --aiwb-only mode

Core Change 
Preserve appDomain parameter in AIWB-only deployments

What Changed:
- Restored appDomain parameter in scripts/init-gitea-job/templates/cf-init-gitea-cm.yaml
- Added back essential configuration:
```
# APP-DOMAIN (preserve from root/values.yaml)
- name: appDomain
  value: {{ .Values.domain | quote }}
```
- Maintained standalone mode with standAloneMode: "true"

Impact
Before:
- AIWB-only deployments missing domain context
- Authentication redirects failing due to undefined domain
- Bootstrap script --aiwb-only flag not preserving domain configuration
After:  
- AIWB standalone deployments inherit domain from root values.yaml
- Proper domain configuration for authentication flows
- Seamless AIWB-only bootstrap functionality restored

Technical Details
- Target Use Case: ./scripts/bootstrap.sh <your-domain> --aiwb-only
- Configuration Source: Domain value templated from .Values.domain 
- Deployment Mode: Standalone AIWB with preserved domain context
- Authentication Flow: Enables proper Keycloak redirect URLs

### DIFF
--- a/scripts/init-gitea-job/templates/cf-init-gitea-cm.yaml
+++ b/scripts/init-gitea-job/templates/cf-init-gitea-cm.yaml
@@ -315,6 +315,9 @@ data:
     cat >> /tmp/values.yaml << 'EOF'
       aiwb:
         helmParameters:
+          # APP-DOMAIN (preserve from root/values.yaml)
+          - name: appDomain
+            value: {{ .Values.domain | quote }}
           # STANDALONE-MODE
           - name: standAloneMode
             value: "true"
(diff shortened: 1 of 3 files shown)


woojae-siloai:
@Q-Dub 
Hi. I had to delete pre-hooks when I tested on pd3. These blocked keycloak argo app.
Every external-secret in keycloak namespace comes from openbao, so I think without pre-hook will be ok.
What do you think?
image

PR #704, opened 2026-05-04
config(cluster-auth): point to 0.5.9 (cache invalidation)

Q-Dub:
config(cluster-auth): point to 0.5.9 (cache invalidation)


woojae-siloai (approved):
LGTM

PR #717, opened 2026-05-26
EAI_5821 evaluate envoy-gateway with node selector

woojae-siloai:
This PR replaces kgateway with envoy-gateway and guarantees that "https" pod of envoy-gateway is deployed on the first node to get "external-ip" for its service.

There is a document how to upgrade cluster using cluster-values: https://github.com/silogen/cluster-forge/blob/EAI_5821_evaluate_envoy_gateway_with_nodeSelector/docs/kgateway-envoy-gateway-migration-guide.md

This is the manifest for upgrading, which gives first-node label and creates "envoy-gateway-system" namespace and copy "cluster-tls" secret from kgateway namespace.

This branch regarding the expected basic functions of envoy-gateway compared to kgateway was tested on app-dev cluster.


brownzebra (approved):
LGTM

PR #723, opened 2026-05-28
moving the init jobs into sources from scripts

brownzebra:


sources/openbao-init-job/0.1.0/values.yaml
```diff
@@ -0,0 +1,19 @@
...
+      enabled: false # from merged config
+  dataStorage:
+    size: 5Gi           # from merged config
+    storageClass: ""    # from merged config
+  resources:            # from merged config
+    requests:
+      cpu: "250m"
+      memory: "512Mi"
+    limits:
+      cpu: "1000m"
```
pre, line 18: CPU limits are an antipattern - they slow down the overall system and cause unnecessary throttling. We shouldn't have CPU limits for any workloads.

pre (approved):
LGTM but the cpu limit should go

PR #729, opened 2026-05-29
feat: custom port for https pod

woojae-siloai:


blankdots (changes requested):
might need to align also the `ai-passthrough` port now

sources/envoy-gateway-config/templates/gateway.yaml
blankdots: port might be needed configure here

blankdots (approved):

PR #730, opened 2026-06-01
EAI-5821: Envoy AI Gateway incremental deployment (4 checkpoints)

johnl-amd:
## Summary

Incremental deployment of Envoy AI Gateway v0.6.0. Each commit is a standalone checkpoint that can be deployed and verified before applying the next.

The previous branch (`EAI-5821-add-envoy-ai-gateway`) deployed everything at once and crashed app-dev-1 because `backendResources: InferencePool` was declared before the CRD existed, putting envoy-gateway into CrashLoopBackOff.

## Deployment checkpoints

**Step 1** — CRDs only, zero envoy-gateway impact
- Installs `inference-extension-crds/v1.5.0` (InferencePool CRD) at syncWave -35
- Installs `envoy-ai-gateway-crds/v0.6.0` (AIGatewayRoute, AIServiceBackend, etc.) at syncWave -10
- ✅ Verify: CRDs visible in cluster, envoy-gateway untouched

**Step 2** — AI Gateway controller, not yet wired to envoy-gateway
- Deploys `envoy-ai-gateway/v0.6.0` controller + webhook at syncWave -5
- envoy-gateway config is unchanged — no xDS impact
- ✅ Verify: `envoy-ai-gateway-system` pods healthy, existing routes unaffected

**Step 3** — Wire extensionManager (envoy-gateway restart required)
- Adds `extensionApis.enableBackend` + `extensionManager` + `backendResources: InferencePool` to envoy-gateway config
- CRDs already exist (step 1) and controller is running (step 2) — no crash
- ✅ Verify: envoy-gateway restarts cleanly, xDS connection to AI Gateway established

**Step 4** — Add AI Gateway token fields to access logs
- Adds `model`, `llm_input_token`, `llm_output_token`, `llm_total_token` to the EnvoyProxy access log format
- Populated by the AI Gateway ext-proc filter for inference requests, `-` for all other traffic
- ✅ Verify: token fields appear in Envoy access logs for AI requests

## Test plan

- [ ] Apply step 1, confirm CRDs installed, envoy-gateway pod unchanged
- [ ] Apply step 2, confirm AI Gateway controller and webhook healthy
- [ ] Apply step 3, confirm envoy-gateway restarts without CrashLoopBackOff
- [ ] Apply step 4, confirm token fields appear in access logs for inference requests


tomastola:
## Step 3 — port 1063 is plain gRPC, not TLS

Tested from inside the cluster with a busybox pod after step 2 deployed:

```
nc -w3 ai-gateway-controller.envoy-ai-gateway-system.svc.cluster.local 1063 | od -c | head -3
0000000  \0  \0 006 004  \0  \0  \0  \0  \0  \0 005  \0  \0   @  \0
```

`\x00\x00\x06\x04` is an HTTP/2 SETTINGS frame — plain gRPC (h2c), no TLS. The `self-signed-cert-for-mutating-webhook` cert is for the mutating webhook on 9443, not the extension server on 1063.

With the `tls:` block present, envoy-gateway will hit the same error we saw on the previous branch attempt:

```
authentication handshake failed: tls: first record does not look like a TLS handshake
```

The fix for step 3: remove the `tls:` block from `extensionManager.service`:

```yaml
extensionManager:
  service:
    host: ai-gateway-controller.envoy-ai-gateway-system.svc.cluster.local
    port: 1063
    # no tls — port 1063 serves plain gRPC (h2c)
```

Steps 1 and 2 verified clean — ai-gateway-controller 1/1 Ready, envoy-gateway 0 restarts in 162 min, all routes returning 200. The incremental structure is solid.

tomastola:
## AI Gateway routing — dedicated Gateway required

The `ai-gateway-discovery` controller was idle because it creates `AIGatewayRoute` objects with `parentRefs` pointing to a Gateway named `ai-gateway`, which didn't exist. Step 5 (just pushed) adds that Gateway.

### Why a dedicated Gateway, not the shared `https` one

The natural question is whether `gateway.app-dev.silogen.ai` could just be another hostname on the existing `https` Gateway, the same way aiwb, airm, keycloak, and every other service works — a plain `HTTPRoute` attached to the shared gateway with Host header matching.

The answer is no, due to how Envoy AI Gateway v0.6.0 works internally. It does **not** use `EnvoyExtensionPolicy`. Instead, it injects the extproc filter as a **sidecar container** directly into Envoy proxy pods via a mutating webhook. The webhook fires on pod creation, checks whether any `AIGatewayRoute` is attached to that pod's Gateway, and if so, injects the sidecar.

This injection is **Gateway-scoped** — it applies to every request handled by those proxy pods, with no mechanism for listener-level or route-level scoping. When we previously pointed `ai-gateway-discovery` at the `https` Gateway, the sidecar was injected into all `https` proxy pods, causing the extproc filter to intercept all traffic including the TLS passthrough listener for `k8s.app-dev.silogen.ai`, which broke it.

A dedicated `ai-gateway` Gateway gets its own separate Envoy proxy pods (Envoy Gateway creates one proxy deployment per Gateway). The sidecar is injected only into those pods. The `https` Gateway and its proxy pods are completely untouched.

### What Step 5 adds

- **`ai-gateway` Gateway** — HTTPS listener on `gateway.app-dev.silogen.ai`, TLS terminated with the same `cluster-tls` wildcard cert, same `access-logging-config` EnvoyProxy, routes allowed from all namespaces
- **`ai-gateway` ClusterIP Service** — selects the `ai-gateway` proxy pods for in-cluster routing (same pattern as the existing `https` ClusterIP Service)
- **CoreDNS rewrite** — exact-match rule for `gateway.app-dev.silogen.ai` → `ai-gateway.envoy-gateway-system.svc.cluster.local`, placed before the wildcard rule with `stop` so it doesn't fall through

The `https` Gateway, TLS passthrough, and all existing services are unaffected.

### External DNS

`gateway.app-dev.silogen.ai` is covered by the existing `*.app-dev.silogen.ai` wildcard DNS record. Envoy Gateway will spin up a separate proxy deployment for `ai-gateway` with its own external LoadBalancer Service — that IP needs a dedicated A record (or the wildcard needs to be adjusted to point at both IPs). Alternatively, the auto-created LoadBalancer can be suppressed in favour of the same external LB used by `https`, with SNI-based routing directing `gateway.app-dev.silogen.ai` traffic to the `ai-gateway` service.

PR #733, opened 2026-06-02
EAI-1500  update Github Workflows documentation

blankdots:


pre (approved):
Sure, why not .. these just need to be kept up-to-date

PR #737, opened 2026-06-08
EAI-5821: Add ext-proc metrics scraping and bump cluster-auth to rc4

johnl-amd:
## Summary

- Adds `ai-gateway-extproc` scrape job to the OTel rest collector — targets Envoy https proxy pods in `envoy-gateway-system` on port 1064 (the ext-proc sidecar admin port) using Kubernetes SD with label filters. No RBAC changes needed; the collector service account already has cluster-wide pod list/watch.
- Bumps `otel-lgtm-stack` v1.0.7 → v1.0.8 to carry the new scrape config
- Bumps `cluster-auth` image tag from 0.6.0-rc3 to 0.6.0-rc4, which adds `x-aim-service-id` header injection via SecurityPolicy contextExtensions

## Test plan

- [ ] OTel metrics-rest collector restarts cleanly with new config (no parse errors in logs)
- [ ] `gen_ai_client_token_usage` metrics appear in Grafana after a model request
- [ ] `x-aim-service-id` header is present on responses proxied through cluster-auth

### DIFF
--- a/sources/cluster-auth/0.5.9/values.yaml
+++ b/sources/cluster-auth/0.5.9/values.yaml
@@ -3,7 +3,7 @@ replicaCount: 1
 image:
   repository: ghcr.io/silogen/cluster-auth
   pullPolicy: Always
-  tag: "0.6.0-rc3"
+  tag: "0.6.0-rc4"
 
 imagePullSecrets: []
 nameOverride: ""
--- a/sources/otel-lgtm-stack/v1.0.7/Chart.yaml
+++ b/sources/otel-lgtm-stack/v1.0.7/Chart.yaml
@@ -8,10 +8,10 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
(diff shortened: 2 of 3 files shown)


tomastola (approved):
Verified against the cluster before reviewing:

- **Port 1064 confirmed correct** — `ai-gateway-extproc` is a Kubernetes sidecar (init container with `restartPolicy: Always`) injected into the Envoy proxy pod, serving Prometheus metrics at `/metrics` on port 1064
- **Label selectors match exactly**: `app.kubernetes.io/name: envoy` and `gateway.envoyproxy.io/owning-gateway-name: https` both present on the target pods
- **Metrics are live**: `gen_ai_client_token_usage` histograms already have real data (tested with a curl to pod IP directly)
- **cluster-auth rc4 bump**: looks good

One minor nit: `sources/otel-lgtm-stack/v1.0.7/Chart.yaml` now declares `version: 1.0.8` but the directory is still named `v1.0.7`. Works fine since ArgoCD renders from Chart.yaml, but future readers hunting for v1.0.8 won't find it intuitively. Worth renaming the directory at some point, not a blocker.

LGTM otherwise.

PR #740, opened 2026-06-09
chore: replace bitnami image references with soldevelo equivalents

jkondrat-sd:
## Why this change

Bitnami moved image updates behind a paywall on August 28 2025, and public images no longer receive ongoing updates.
[SolDevelo](https://hub.docker.com/u/soldevelo) images are free drop-in replacements with active support.

## Image replacements

- `bitnami/kubectl:latest` → [`soldevelo/kubectl:latest`](https://hub.docker.com/r/soldevelo/kubectl)

### DIFF
--- a/scripts/utils/envoy-gateway-migration.yaml
+++ b/scripts/utils/envoy-gateway-migration.yaml
@@ -179,7 +179,7 @@ spec:
       # Phase 2: Node Labeling (Main Container)
       containers:
       - name: node-labeling
-        image: bitnami/kubectl:latest
+        image: soldevelo/kubectl:latest
         command: ["/bin/bash"]
         args:
         - -c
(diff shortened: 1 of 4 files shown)


scripts/utils/envoy-gateway-migration.yaml
```diff
@@ -179,7 +179,7 @@ spec:
       # Phase 2: Node Labeling (Main Container)
       containers:
       - name: node-labeling
-        image: bitnami/kubectl:latest
+        image: soldevelo/kubectl:latest
```
pre, line 182: Removing the Bitnami dependency is reasonable, but would `alpine/kubectl` or similar distro based source work instead?

jkondrat: That's a valid alternative. I suggested SolDevelo because it's a Bitnami-compatible drop-in replacement, which keeps the change minimal and avoids additional testing or configuration changes beyond replacing the image source.

PR #750, opened 2026-06-16
fix: point to the correct keykloak devuser secret

blankdots:
related to https://github.com/silogen/cluster-bloom/pull/262


blankdots:
not needed deprecated docs

PR #760, opened 2026-06-22
add ai-gateway-discovery app, and add envoy/ai-gateway overrides to root/values.yaml

oskarasbrink:
add ai-gateway-discovery app, and add envoy/ai-gateway overrides to root/values.yaml

### DIFF
--- a/root/values.yaml
+++ b/root/values.yaml
@@ -16,14 +16,16 @@ ociRegistry:
 apps:
   ai-gateway-discovery:
     repoURL: "{{ .Values.ociRegistry.dockerHub }}"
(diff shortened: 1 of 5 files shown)


blankdots (changes requested):
lgtm, though might want to keep the sbom components up to date as well

```
components:
  ai-gateway-discovery:
    path: null
    repoVersion: 2.0.0
    valuesFile: values.yaml
    sourceUrl: <url>
    projectUrl:  <url>
    license: License
    licenseUrl:  <url>
```

blankdots (approved):

PR #764, opened 2026-06-24
EAI-7043: fix UUID-based AIM routing via Lua filter

johnl-amd:
## Summary

- Lua filter (`set-model-header-from-body`) now skips setting `x-ai-eg-model` when a `backend` UUID is present in the request body or headers, so model-name fallback rules in other HTTPRoutes cannot match. The UUID rule (Host + `x-ai-eg-backend`) wins by specificity over catch-all rules regardless of HTTPRoute creation order.
- Introduces `x-ai-eg-model-log` header, always set from the body's `model` field, so the access log retains model attribution for both UUID-routed and model-name-routed requests.
- Access log format updated to read `x-ai-eg-model-log` instead of `x-ai-eg-model`.

## Root cause

Envoy Gateway merges HTTPRoute rules maintaining per-route creation-timestamp ordering, not global specificity. Two rules with equal header-match counts from different HTTPRoutes — the older route always wins. When two AIMs serve the same model, the first-created route's model-name rule would always win, routing to the wrong AIM regardless of the `x-ai-eg-backend` header.

## Test plan

- [ ] Deploy to app-dev and send a request with `"backend": "<uuid>"` in the body; verify `backendCluster` in access log shows the correct AIM's HTTPRoute rule
- [ ] Verify `model` field in access log is still populated for UUID requests
- [ ] Verify model-name routing (no `backend` field) still works for standard OpenAI clients

### DIFF
--- a/sources/cluster-auth/0.5.9/values.yaml
+++ b/sources/cluster-auth/0.5.9/values.yaml
@@ -3,7 +3,7 @@ replicaCount: 1
 image:
   repository: ghcr.io/silogen/cluster-auth
   pullPolicy: Always
-  tag: "0.6.0-rc8"
+  tag: "0.6.0-rc9"
 
 imagePullSecrets: []
 nameOverride: ""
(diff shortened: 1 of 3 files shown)


blankdots (approved):
lgtm

PR #767, opened 2026-07-01
EAI-5630 again aiwb only fix

woojae-siloai:
This PR adds "appDomain" helmparameter at [cf-init-gitea-cm.yaml](https://github.com/silogen/cluster-forge/compare/EAI-5630-again-aiwb-only-fix?expand=1#diff-216b03f7c0ea4f8dcfed2c57ac8dd1f72e4c302704f6cd8b7335215c4d241355).

In short, this will work if there is a proper "2.0.0" version of aiwb-chart pointing envoy-gateway-system for its httproutes.

It is tested on a test machine with modification of
- aiwb.repoVersion: "1.1.9" at root/values.yaml
- at gitea's cluster-values as follows
```
aiwb:
  helmParameters:
    # KGATEWAY-NAMESPACE
    - name: kgateway.namespace
      value: "envoy-gateway-system"
```

aiwb.repoVersion: "2.0.0-rc.1"  doesn't exist so "1.1.9" selected as a workaround to run aiwb. Plus "1.1.9" versions uses "kgateway.namespace: kgateway-system" so updating cluster-values was needed for testing.

The AIWB Helm chart defaults to `appDomain: "PUBLIC-IP"`, but ArgoCD Application helmParameters (from cluster-values.yaml) have **highest priority** and override chart defaults. Without adding `appDomain` to helmParameters in gitea-init-job, the deployed AIWB uses the placeholder value, causing NEXTAUTH_URL to be `https://aiwbui.PUBLIC-IP` instead of the actual domain. This fix ensures gitea-init-job generates cluster-values.yaml with the correct `appDomain` override for AIWB_ONLY deployments.

### DIFF
--- a/root/values.yaml
+++ b/root/values.yaml
@@ -67,7 +67,7 @@ apps:
         kind: ClusterPolicy
   aiwb:
     repoURL: "{{ .Values.ociRegistry.dockerHub }}"
-    repoVersion: "2.0.0-rc.1"
+    repoVersion: "1.1.9"
     chart: "aiwb-chart"
     namespace: aiwb
     valuesFile: values.yaml
--- a/root/values.yaml
+++ b/root/values.yaml
@@ -67,7 +67,7 @@ apps:
         kind: ClusterPolicy
   aiwb:
     repoURL: "{{ .Values.ociRegistry.dockerHub }}"
-    repoVersion: "1.1.9"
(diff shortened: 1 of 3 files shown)


pre (approved):
Nice! I was also hit by this yesterday.

PR #775, opened 2026-07-10
Add option to use local Helm values for apps

wtripp180901:
Currently the only way override Helm values in a way that the changes won't be reverted by ArgoCD is to use externalValues and reference an external repository or the local Gitea server containing an overrides files. For local development or when consuming cluster-forge dependencies as part of a manual Helm installation of AIWB/AIRM where a Gitea server is not a hard dependency, it is useful to be able to iterate on the configuration purely through upgrading the Helm release rather than pushing commits to a remote repository.

Adds `localHelmValues.enabled` option which disables the cluster-forge wrapper Application. This allows the user to deploy cluster-forge with local overrides e.g with `helm install cluster-forge ./root --values values_<size>.yaml --values custom_overrides.yaml` while ArgoCD still consumes the remote OCI artifacts and defaults for each of the constituent applications for a given revision of cluster-forge.

Needs docs but the docs/configuration-reference.md files referenced in CONTRIBUTING.md seems to be missing so unsure where the best place for these would be.


brownzebra (approved):
Looks good to me

wtripp180901:
Superseded by https://github.com/silogen/cluster-forge/pull/787

PR #776, opened 2026-07-13
docs: update main readme

blankdots:
- update first deployment installed apps
- boostrap.sh is not present any more so update accordingly


pwistbac (approved):
Great job

PR #780, opened 2026-07-17
fix via feature: add retry policy to cluster-apps.yaml template

Q-Dub:
fix via feature: add retry policy to cluster-apps.yaml template
additionally add argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true (clean retry over hard error)


blankdots (approved):
lgtm

PR #790, opened 2026-07-27
EAI-7555_Overhaul_lgtm_stack

woojae-siloai:
Switches to the **official Grafana image** and lets you **configure the LGTM stack per cluster** by editing `values.yaml` — no image rebuild.

## Why
On small clusters the `lgtm` pod kept crashing (out of memory) and the Loki disk filled up. The fixes were **hardcoded** before, so you couldn't adjust them per cluster. Now you can.

## What you can configure now
In `sources/otel-lgtm-stack/v1.0.8/values.yaml`:

- **Memory** — how much memory the apps aim to use (`GOMEMLIMIT`).
- **Retention** — how long to keep Prometheus metrics and Loki logs.
- **Scrape intervals** — how often metrics are collected (global + per job).
- **Collector memory guard** — when collectors start refusing data.
- **Grafana login** — auth settings.

To change a value for one cluster, override it in that cluster's `cluster-values.yaml`.

##   Two ways to configure — don't mix them up
1. **Individual parameters** (most settings) — change **only the one line** you want. Everything else keeps its default.
2. **Whole config files** (`lokiConfig`, `otelcolConfig`) — these are one big block. To change even one line, you must **copy the entire block**.

Also: `PROMETHEUS_EXTRA_ARGS` must use YAML folded style (`>-`), not `|` (otherwise flags get dropped).

📖 Full guide with copy-paste samples: **[`docs/otel_lgtm_stack_tuning.md`](https://github.com/silogen/cluster-forge/blob/overhaul_lgtm_stack_v3/docs/otel_lgtm_stack_tuning.md)**

## Files to be highlighted
- **`values.yaml`** — every setting you can configure (start here).
- **`lgtm-stack.yaml`** — the official image tag (`grafana/otel-lgtm:0.29.2`).
- The 3 `collectors-*.yaml` files just pass values through — quick skim.

## Validation
Tested end-to-end on a live cluster: a `cluster-values.yaml` override flowed through gitea → ArgoCD → pod restart → running config. All settings applied.

## Out of scope
- Filtering out noisy metrics/logs — separate PR.
- Giving each app its own memory budget — needs image changes; `GOMEMLIMIT` is a shared limit that every app inherits equally.

### DIFF
--- a/docs/otel_lgtm_stack_tuning.md
+++ b/docs/otel_lgtm_stack_tuning.md
@@ -1,0 +1,145 @@
+# Tuning the otel-lgtm-stack per cluster
+
+This chart runs the full **LGTM** observability stack (Grafana + Loki + Prometheus
++ Tempo + Pyroscope) inside a single `lgtm` pod, plus a set of OpenTelemetry
+**collectors**. On small / under-specced clusters the `lgtm` pod (and some
+collectors) can hit **OOM**, and the Prometheus / Loki data volumes can fill up.
+
+This refactor exposes the knobs needed to deal with that **from
+`cluster-values.yaml` in gitea** — no template edits and no image rebuild.
+
+---
+
+## TL;DR — what you can tune, and how
+
+| What you want to change            | Override path                                   | How much to write            |
+| ---------------------------------- | ----------------------------------------------- | ---------------------------- |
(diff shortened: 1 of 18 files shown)


oskarasbrink (approved):
Awesome Woojae, the docs are a nice addition also 💯

PR #791, opened 2026-07-28
feat: extraApps for argo

lauri-amd:
Adds a pluggable "extra components" mechanism: a separate cluster-forge-extras app-of-apps (root-extras/ chart) lets users install arbitrary ArgoCD Applications via an extraApps map in their cluster-values overlay, without touching or forking the core cluster-forge stack.
Each extraApps.<name> entry is a raw ArgoCD Application spec: block, so any Application.spec feature is supported; entries live in a dedicated extra-apps-values.yaml file to keep them isolated from the main overlay values.yaml


blankdots (approved):
lgtm

PR #792, opened 2026-07-29
feat(extra-apps): Add blueprint for vllm-sr

lauri-amd:
Example blueprint for adding vllm-sr as an extra app. The config for it is just a placeholder for demonstrating the idea.

The user is required to just use these as reference and implement these in the per cluster cluster-values repo.


brownzebra (approved):
Great, thank you.

PR #796, opened 2026-08-04
Eai 7068 upgrade seaweedfs

pwistbac:
Upgrade seaweedfs to 0.1.36


sources/seaweedfs-operator/0.1.36/values.yaml
```diff
@@ -0,0 +1,189 @@
...
+    # can inject the CA into the webhook configurations; CA, Vault, and
+    # self-signed issuers do, ACME issuers such as Let's Encrypt do not.
+    issuerRef: {}
+
+## seaweedfs-operator containers' resource requests and limits.
+## ref: http://kubernetes.io/docs/user-guide/compute-resources/
+resources:
+  limits:
+    # -- seaweedfs-operator containers' cpu limit (maximum allowed CPU)
+    cpu: 500m
```
pre, line 153: In general I think we should not have CPU limits anywhere. CPU limits have the risk of slowing down the system overall, instead of protecting for resource hogs.

pre (approved):
OK if it works, but I'd remove any CPU limits.

PR #801, opened 2026-08-07
EAI-7854 Pin aiwb-infra-cnpg to one instance on small profile

pre:
Related:
* https://github.com/silogen/cluster-forge/pull/798

The `small` profile did not override `instances` for `aiwb-infra-cnpg`, so the
chart default (multiple replicas) was used on single-node clusters where the
extra CNPG replicas cannot be scheduled.

Set `instances: 1`, matching what `root/values_medium.yaml` already does.

## Testing

Renders on the small profile with a single CNPG instance for `aiwb-infra-cnpg`.

### DIFF
--- a/root/values_small.yaml
+++ b/root/values_small.yaml
@@ -60,6 +60,9 @@ enabledApps:
   - rabbitmq
 
 apps:
+  aiwb-infra-cnpg:
+    valuesObject:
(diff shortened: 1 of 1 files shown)


silokimmo (approved):
LGTM

PR #807, opened 2026-08-12
EAI-5958: Add kubeletstats collector to stamp workload_id on pod CPU/mem metrics

johnl-amd:
> **Note on base branch:** this targets `bump_version_envoy_and_ai_gateway` (the gateway/ai-gateway bump currently deployed on app-dev) rather than `main`, so the collector lands on top of that branch without disturbing the gateway stack app-dev is running. Supersedes #806 (same change, based on `main`).

## Summary
- Adds a dedicated `kubeletstats` daemonset collector (`otel-collector-metrics-kubelet`) to the `otel-lgtm-stack` chart that emits **per-pod CPU and memory** metrics labeled with **`workload_id`** (plus `project_id`, `pod`, `namespace`, `container`).
- Enables per-workload CPU/system-memory observability for EPYC (CPU-only) AIMs in the AIWB dashboard, scoped the same stable way GPU/VRAM metrics already are.

## Why
EPYC-based AIMs (parent initiative EAI-2244, epic EAI-5958) hold model weights in RAM, so the AIWB dashboard needs CPU + system-memory panels alongside the existing GPU/VRAM panels. The AIWB metrics backend scopes every panel by a `workload_id="<AIM-uuid>"` label. GPU metrics carry that label (stamped by the device-metrics-exporter); vLLM metrics carry it (vLLM collector relabel). **Container CPU/memory metrics do not.**

## Root cause of the gap
`otel-collector-metrics-k8s` scrapes cAdvisor via a `role: node` scrape (`/metrics/cadvisor`). Those metrics are **node-scoped**, so `pod`/`namespace` arrive as *datapoint* labels. The `k8sattributes` processor associates pods at the **resource** level, so it cannot enrich node-scoped cAdvisor series — leaving `container_cpu_usage_seconds_total` / `container_memory_working_set_bytes` without `workload_id`.

## Approach
Use the `kubeletstats` receiver instead, which emits **pod/container-scoped** metrics with `k8s.pod.*` as resource attributes:
1. `kubeletstats` (daemonset, per-node kubelet at `:10250`) → pod + container metric groups. The kubelet is addressed by **node IP** (`status.hostIP`), not node name — this cluster's CoreDNS doesn't resolve node hostnames, so `${env:K8S_NODE_NAME}:10250` fails with `no such host`.
2. `k8sattributes` looks up the pod and extracts `airm.silogen.ai/workload-id` → `workload.id` (and `project-id`).
3. `transform/workload-id-labels` copies the identity resource attributes **down to datapoint attributes**, so they survive OTLP→Mimir ingestion as **series labels** (resource attributes otherwise land in `target_info`, not on the series).

Reuses the existing `otel-collector` ServiceAccount — RBAC already grants `nodes/stats`, `nodes/proxy`, `pods`, `namespaces`, so no RBAC change is needed.

## Risk
Low–medium. Additive: a new collector; existing pipelines are untouched. New series (`k8s_pod_cpu_*`, `k8s_pod_memory_*`, `k8s_container_*`) with per-pod cardinality (comparable to existing cAdvisor series). Uses the same collector image and OTLP endpoint as the other collectors.

## Test plan / validation status — ✅ validated on app-dev (2026-08-12)
Validated end-to-end against existing labeled pods (no EPYC model needed — the collector stamps `workload_id` on any pod carrying `airm.silogen.ai/workload-id`):

- Collector scrapes cleanly on all 4 nodes after the node-IP fix (no `no such host`).
- `k8s_pod_memory_working_set_bytes{workload_id="<aim-uuid>"}` returns a series carrying `workload_id`, `project_id`, `namespace`, `pod` as labels.
- `k8s_pod_cpu_utilization_ratio{workload_id="<aim-uuid>"}` returns the pod's CPU ratio.
- 4/4 running labeled pods on app-dev show a non-empty `workload_id` on their pod memory series.

**Emitted metric names** (note the OTLP→Prometheus unit suffixes, for the AIWB backend follow-up):

| Purpose | Metric |
|---|---|
| CPU usage | `k8s_pod_cpu_utilization_ratio` (ratio → ×100 for %), `k8s_pod_cpu_time_seconds_total` (counter) |
| System memory usage | `k8s_pod_memory_working_set_bytes`, `k8s_pod_memory_usage_bytes` |
| System memory available | `k8s_pod_memory_available_bytes` |

## Follow-ups (separate work, not in this PR — the rest of epic EAI-5958)
- AIWB backend: add `MetricName` entries + PromQL builders for CPU usage / CPU availability / system memory, using the **actual emitted names above** (`k8s_pod_cpu_utilization_ratio`, `k8s_pod_memory_working_set_bytes`, …).
- AIWB frontend: CPU/memory panels + accelerator-type gating (CPU usage panel, CPU availability panel, system-memory panel alongside the GPU panel).
- CPU/memory **availability** denominator sourced from the K8s object (resolved-profile requests/limits) rather than kube-state-metrics — kubeletstats does not emit the pod's CPU limit.


johnl-amd:
### Why collect CPU/memory this way (kubeletstats + k8sattributes)

Recording the design rationale, since the approach isn't obvious.

**Requirement:** AIWB needs per-pod CPU/memory series **natively labelled with `workload_id`**, so the backend queries them with the same `{workload_id="..."}` selector it already uses for GPU and vLLM metrics. Keeping the query layer uniform is the goal.

**Why not reuse the existing cAdvisor scrape?** We already scrape `container_cpu_*`/`container_memory_*` via the `kubernetes-nodes-cadvisor` job, but those come in through the `node` role: `pod`/`namespace` are *datapoint* labels and there's no `workload_id`. The `k8sattributes` processor associates pods at the *resource* level, and cAdvisor series carry no `k8s.pod.*` resource attributes — so it **cannot enrich them**. cAdvisor + enrichment is a dead end.

**Why not a query-time join?** We could keep cAdvisor and join against `kube_pod_labels` (kube-state-metrics) in every CPU/mem query. That avoids a new collector, but makes CPU/mem a special case: every AIWB query becomes a fragile PromQL join, diverging from the clean `{workload_id=...}` pattern used everywhere else and depending on kube-state-metrics label exposure. Rejected for consistency/maintainability.

**Why not source-side stamping (like GPU/vLLM)?** GPU works because the device-metrics-exporter knows which workload owns each GPU and labels at emission; vLLM relabels its own metrics. Generic CPU/memory comes from the kernel/cgroups via the kubelet, which is workload-agnostic — there's no app to stamp `workload_id` at the source. So an enrichment step is unavoidable.

**Chosen approach:** `kubeletstats` emits pod-scoped metrics with `k8s.pod.*` as *resource* attributes → `k8sattributes` looks up the pod and adds `workload_id`/`project_id` → a transform promotes them to datapoint labels so they land as Prometheus series labels (not `target_info`). Result: `k8s_pod_cpu_utilization_ratio{workload_id="..."}`, queried identically to GPU/vLLM.

**Cost & mitigations:** this path needs a DaemonSet (each node's kubelet is only reachable locally) and enrichment (hence a pod watch) — costs GPU/vLLM avoid by getting the label at the source. Mitigated here by: node-scoping the `k8sattributes` watch (`node_from_env_var`, so cost scales pods-per-node not pods×nodes), dropping the unused `container` metric group (pod-level is all AIWB queries), and reusing the existing least-privilege `otel-collector` ServiceAccount.

Not prescribed by EAI-5958 (the epic specifies the outcome, not the mechanism) — this is the mechanism that keeps CPU/memory consistent with the existing GPU/vLLM metric shape.

nowycondro (approved):

brownzebra:
@johnl-amd Can you guys give any testing evidence here?

PR #808, opened 2026-08-12
Envoy collector scale to zero

williamanzen:
# Summary

- Add Gateway-scoped Lua activation counters for AIM scale-from-zero, running after external authorization.
- Add a dedicated OpenTelemetry collector that forwards normalized delta metrics to KEDA.
- Wire the collector into all cluster profiles and SBOM metadata.
- Document the temporary InsecureSyntax risk and restrict EnvoyExtensionPolicy management to platform administrators.

## Security
InsecureSyntax is enabled only for scale-from-zero because Envoy Gateway’s strict validator does not yet support handle:stats(). Lua runs after ext_authz, and platform-admin RBAC limits policy mutation.

## Test plan

- [x]  Helm lint and render changed charts.
- [x]  Render small, medium, and large profiles.
- [x]  Validate scale-from-zero with Envoy Gateway 1.8.2.
- [x]  Confirm the 1.8.1 deployment uses Envoy 1.38.1 
- [ ]  rerun the integration test - E2E testing of scale-from-zero on 1.8.1

PR #810, opened 2026-08-17
EAI-5958: Add kubeletstats collector to stamp workload_id on pod CPU/mem metrics

johnl-amd:
## Summary
Adds the EAI-5958 **kubeletstats collector** that stamps `workload_id` (and `project_id`) onto per-pod CPU/memory metrics, so per-workload CPU/mem observability matches how GPU and vLLM metrics already carry `workload_id`. Cherry-picked from #808 (still open) and ported to the deployed chart version.

## Why a dedicated kubeletstats daemonset
The cAdvisor metrics scraped by `otel-collector-metrics-k8s` are **node-scoped**, so pod/namespace arrive as datapoint labels and can't be associated to a pod by the `k8sattributes` processor (which works at the resource level). This daemonset uses the **kubeletstats** receiver, which emits pod/container-scoped metrics with `k8s.pod.*` as resource attributes — so `k8sattributes` can look up the pod and stamp its `airm.silogen.ai/workload-id` label. A `transform` then copies those identity resource attributes down to datapoint attributes so they survive OTLP ingestion as series labels (`workload_id`, `project_id`, `pod`, `namespace`, `container`).

## Contents
- **`v1.0.7`** — the four original EAI-5958 commits from #808, cherry-picked unchanged (kubeletstats collector; kubelet addressed by node IP `status.hostIP` not node name; review fixes — pod-only metric group + `tolerations: Exists`; node-scoped `k8sattributes` watch).
- **`v1.0.8`** — the same collector ported to the **active** version. `root/values.yaml` deploys `otel-lgtm-stack/v1.0.8`, so the `v1.0.7` change alone would be inert (as it is on #808 today). The port wires `memory_limiter` and the kubelet `collection_interval` to the `.Values` keys v1.0.8 parameterized (`collectors.memoryLimiter.*`, `collectors.scrapeInterval`), matching every other collector in that chart.

## Non-obvious decisions
- **Kubelet by node IP, not node name** — cluster DNS (CoreDNS) does not resolve node hostnames here, so `${env:K8S_NODE_NAME}:10250` fails with "no such host"; `status.hostIP` is always routable. `K8S_NODE_NAME` is kept only as the local `k8sattributes` watch filter (not for DNS).
- **Node-scoped `k8sattributes` watch** — each daemonset pod watches only its own node's pods (`node_from_env_var: K8S_NODE_NAME`), so cache cost scales with pods-per-node instead of pods × nodes.
- **Pod-level metric group only** — the container group would roughly double per-container cardinality (overlapping the existing cAdvisor scrape) for series nothing reads.
- **`tolerations: Exists`** — runs on every node including tainted GPU/inference nodes, so per-workload CPU/mem is collected fleet-wide (matches the node-exporter daemonset's scheduling).

## Test plan
- `helm template sources/otel-lgtm-stack/v1.0.8` renders clean (exit 0); parameterized values substitute correctly in the new collector.
- `helm lint sources/otel-lgtm-stack/v1.0.8` passes.
- Deploy verification: confirm the `otel-collector-metrics-kubelet` daemonset runs on every node and that `k8s_pod_*` CPU/memory series carry the `workload_id` / `project_id` labels in Grafana/LGTM.

### DIFF
--- a/sources/otel-lgtm-stack/v1.0.7/templates/collectors-logs-metrics-k8s.yaml
+++ b/sources/otel-lgtm-stack/v1.0.7/templates/collectors-logs-metrics-k8s.yaml
@@ -723,3 +723,111 @@ spec:
           receivers: [prometheus]
           processors: [memory_limiter, batch, attributes]
           exporters: [otlp]
+---
+# Per-pod CPU / memory metrics enriched with workload_id.
+# cAdvisor metrics scraped by otel-collector-metrics-k8s are node-scoped, so their
+# pod/namespace arrive as datapoint labels and cannot be associated to a pod by the
+# k8sattributes processor (which works at the resource level). This daemonset uses the
+# kubeletstats receiver, which emits pod/container-scoped metrics with k8s.pod.* as
+# resource attributes, so k8sattributes can look up the pod and stamp its
+# airm.silogen.ai/workload-id label. A transform then copies the identity resource
+# attributes down to datapoint attributes so they survive OTLP ingestion as series
+# labels (workload_id, project_id, pod, namespace, container) — matching how GPU and
+# vLLM metrics already carry workload_id for per-workload observability.
+apiVersion: opentelemetry.io/v1beta1
+kind: OpenTelemetryCollector
+metadata:
+  name: otel-collector-metrics-kubelet
+  namespace: {{ .Release.Namespace }}
+spec:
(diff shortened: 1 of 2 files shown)


nowycondro (approved):

woojae-siloai (approved):
LGTM

PR #813, opened 2026-08-18
fix: TOCTOU race in openbao secret-manager script

lauri-amd:
The secret-manager job classified an unreachable/sealed vault the same as "secret does not exist" (both a non-zero `bao kv get`), so it could overwrite a live random secret if it raced the unseal job. Distinguish those cases explicitly and make writes self-defending instead of trusting the read:

- kv_status() classifies existence by HTTP status from /v1/secrets/metadata/ (matches what -cas checks) instead of a plain CLI exit code, with connect/max timeouts so a hung vault can't wedge the job
- random secrets are written with -cas=0, so even a wrong "missing" read gets rejected server-side instead of clobbering
- apply the same status-based check + CAS guard to the AIWB scoped token mint/rotation, which had the identical blind-overwrite bug
- guard envsubst substitutions and count/report failures instead of swallowing them under set -e, so the job actually exits non-zero on partial failure
- offset the cronjob schedule from the unseal job's to shrink the overlap window (defense in depth; -cas=0 is the real guard)
- give the init Job an activeDeadlineSeconds, since it can now legitimately exit non-zero where it previously always exited 0

### DIFF
--- a/sources/openbao-config/0.1.0/templates/openbao-secret-manager-cm.yaml
+++ b/sources/openbao-config/0.1.0/templates/openbao-secret-manager-cm.yaml
@@ -6,38 +6,79 @@ metadata:
 data:
   manage-secrets.sh: |
     #!/bin/bash
-    set -e
-    
+    set -euo pipefail
+
     # Validate environment
-    if [ -z "$BAO_TOKEN" ]; then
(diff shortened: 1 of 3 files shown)


Q-Dub (approved):
validated via test plan in Jira: https://amd.atlassian.net/browse/EAI-8139

PR #815, opened 2026-08-21
EAI-8229 Fix CLUSTER_SIZE small: keep global.domain, hooks, drop gitea

pre:
Fixes [EAI-8229](https://amd.atlassian.net/browse/EAI-8229).

`CLUSTER_SIZE: small` does not give a platform that works. Three faults stop
it, one after the other. This pull request corrects all three.

| # | Fault | Correction |
|---|---|---|
| 1 | The root application loses `global.domain`, and CoreDNS then sends every name to the gateway | Keep the value as a parameter |
| 2 | `envoy-gateway` never leaves PreSync, so the Gateway API CRDs are never installed | Give the hooks a delete policy |
| 3 | `gitea` waits for a secret that nothing makes on this path | Do not enable it on the small path |

Only the small path is affected. With `externalValues.enabled`, the values of
fault 1 come from `cluster-values/values.yaml` in the in-cluster Gitea.

---

## 1. The root application loses global.domain

The root Application is rendered two times. bloom renders it with
`helm template ... --set global.domain=...` and applies the result. ArgoCD then
renders the same chart again to manage the app-of-apps, and that render reads
only the value files.

`root/templates/cluster-forge.yaml` copied `clusterForge.targetRevision` into
`spec.source.helm.parameters`, but not `global.domain` and not
`global.clusterSize`. The second render saw them empty, and the root
application overwrote its own specification without them.

An empty domain made `envoy-gateway-config` write

```
rewrite continue name regex .*\. https.envoy-gateway-system.svc.cluster.local answer auto
```

`.*\.` matches every name:

```
argocd-repo-server.argocd.svc.cluster.local   -> 10.243.148.243
argocd-redis.argocd.svc.cluster.local         -> 10.243.148.243
nope-does-not-exist.default.svc.cluster.local -> 10.243.148.243
```

`10.243.148.243` is the `envoy-gateway-system/https` service. The ArgoCD
controller could not reach its repo-server, and all 39 applications stayed in
`ComparisonError`.

An empty `global.clusterSize` also removed `values_small.yaml` from the value
files, so the cluster got the wrong set of applications. The count decreased
from 39 to 26.

**Correction.** Copy both values into the parameters, in the same way as
`clusterForge.targetRevision`. The render now gives the same result each time:

```
helm template cluster-forge root --show-only templates/cluster-forge.yaml \
  --values root/values.yaml --values root/values_small.yaml \
  --set global.domain=1.2.3.4.nip.io | grep -A1 global.domain
```

## 2. envoy-gateway never leaves PreSync

```
waiting for completion of hook /ServiceAccount/envoy-gateway-gateway-helm-certgen and 5 more hooks
```

The hooks in `sources/envoy-gateway/v1.8.1` carried `helm.sh/hook` but no
`helm.sh/hook-delete-policy`. ArgoCD reads a missing policy as
`BeforeHookCreation`, so it deletes each hook before it makes the hook again,
and it does that each time it resumes the operation. The operation resumes
about every four seconds, and the certgen job needs more time than that, so
the job never started.

Measured: the operation was `Running` for 30 minutes, and
`envoy-gateway-system` held no service account, no job and no pod. The Gateway
API CRDs were never installed, so each application that uses a `Gateway`, an
`HTTPRoute` or a `TLSRoute` failed with

```
failed to discover server resources for group version gateway.networking.k8s.io/v1
```

Ten applications stayed `OutOfSync/Missing`, among them `envoy-gateway-config`,
`keycloak`, `gitea-config`, `openbao-config` and `seaweedfs-config`.

**Correction.** `hook-delete-policy: hook-succeeded` on the seven hooks of that
chart. The hooks now stay until the job completes.

## 3. gitea cannot start on the small path

bloom does not bootstrap gitea when `CLUSTER_SIZE` is small.
`clusterforge_setup.yaml` holds `when: (CLUSTER_SIZE | default('medium')) !=
'small'`, and `bootstrap_gitea.yaml` is the only thing that makes the secret
`gitea-admin-credentials`. `values_small.yaml` enabled the `gitea` and
`gitea-config` applications all the same.

Measured: `secret "gitea-admin-credentials" not found`, 255 restarts in 59
minutes, and the application stayed `Degraded`.

**Correction.** Do not enable the two applications on the small path. That path
reads the repository from GitHub, because `externalValues.enabled` is false and
`clusterForge.repoUrl` names GitHub, so it has no use for the in-cluster Gitea.
Nothing else in `values_small.yaml` names it.

---

## Test

One Kaytoo VM, Ubuntu 24.04, no GPU, `CLUSTER_SIZE: small`,
`DOMAIN: <ip>.nip.io`, `CERT_OPTION: generate`. Installed with bloom, then
cluster-forge from this branch.

| | Before | After |
|---|---|---|
| root app parameters | `clusterForge.targetRevision` only | + `global.domain`, `global.clusterSize` |
| root app valueFiles | `values.yaml` | `values.yaml`, `values_small.yaml` |
| applications | 26 | 53 |
| `ComparisonError` | 39 of 39 | 0 |
| not `Synced+Healthy` | 39 | 1 |

Name resolution now goes both ways, as intended:

```
regex .*\.129.213.21.55.nip.io

argocd-repo-server.argocd.svc.cluster.local  -> 10.243.9.71      (the service)
kc.129.213.21.55.nip.io                      -> 10.243.209.223   (the gateway)
```

envoy-gateway installs and both gateways are programmed:

```
phase: Succeeded, message: successfully synced (all tasks run)
pod/envoy-gateway-5f775fc774-tl7bs   1/1   Running

ai-gateway   envoy-gateway   10.243.109.13   True
https        envoy-gateway   10.0.255.153    True
```

## What stays open

One application, `envoy-gateway-config`, is still `OutOfSync`, and it is not
corrected here. `security-policy-ai-gateway-default-deny.yaml` asks for port
8083 on `ai-gateway-system/ai-gateway-discovery`, and envoy-gateway refuses the
policy:

```
ExtAuth: TCP Port 8083 not found on service ai-gateway-system/ai-gateway-discovery
```

`ai-gateway-discovery-chart:2.0.0` gives one port, `health:8081`, and the
container gives only `health:8081`, so there is no ext-auth port. This is a
mismatch of versions and not a wrong number, and port 8081 is the health probe,
so no correction is guessed here.
[EAI-8231](https://amd.atlassian.net/browse/EAI-8231) records it.

[EAI-8229]: https://amd.atlassian.net/browse/EAI-8229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EAI-8231]: https://amd.atlassian.net/browse/EAI-8231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/root/templates/cluster-forge.yaml
+++ b/root/templates/cluster-forge.yaml
@@ -31,10 +31,29 @@ spec:
     targetRevision: {{ .Values.clusterForge.targetRevision }}
     path: root
     helm:
-      {{- if .Values.clusterForge.targetRevision }}
+      # Every value bloom passed with --set when it first rendered this file
+      # must be repeated here. ArgoCD renders this same chart again to manage
+      # the app-of-apps, and that render sees only the valueFiles below, so a
+      # value that is not a parameter is lost on the first sync and the root
+      # app overwrites its own spec without it. global.domain went missing
+      # this way: envoy-gateway-config then wrote a CoreDNS rewrite of
+      # `.*\.` and every name in the cluster resolved to the gateway.
+      # externalValues.enabled carries these in cluster-values/values.yaml
+      # instead, which is why only the small path was affected.
+      {{- if or .Values.clusterForge.targetRevision .Values.global.domain .Values.global.clusterSize }}
       parameters:
+        {{- if .Values.clusterForge.targetRevision }}
         - name: clusterForge.targetRevision
           value: {{ .Values.clusterForge.targetRevision | quote }}
+        {{- end }}
+        {{- if .Values.global.domain }}
+        - name: global.domain
+          value: {{ .Values.global.domain | quote }}
+        {{- end }}
+        {{- if .Values.global.clusterSize }}
+        - name: global.clusterSize
+          value: {{ .Values.global.clusterSize | quote }}
+        {{- end }}
       {{- end }}
       valueFiles:
         - {{ .Values.clusterForge.valuesFile }}
--- a/root/values_small.yaml
+++ b/root/values_small.yaml
@@ -31,8 +31,6 @@ enabledApps:
   - envoy-gateway-config
   - envoy-ai-gateway-crds
   - envoy-ai-gateway
-  - gitea
-  - gitea-config
   - kaiwo
   - kaiwo-config
   - kaiwo-crds
--- a/sources/envoy-gateway/v1.8.1/templates/certgen-rbac.yaml
+++ b/sources/envoy-gateway/v1.8.1/templates/certgen-rbac.yaml
@@ -10,6 +10,12 @@ metadata:
   {{- end }}
   annotations:
     "helm.sh/hook": pre-install, pre-upgrade
+    # ArgoCD deletes a hook that has no delete policy before it makes the
+    # hook again, and it does that on each resume of the operation. The
+    # certgen job needs more time than the interval between two resumes,
+    # so the job never started and envoy-gateway stayed at PreSync. Keep
+    # the hook until it succeeds.
+    "helm.sh/hook-delete-policy": hook-succeeded
     "helm.sh/hook-weight": "-1"   # Ensure rbac is created before the certgen job when using ArgoCD.
   {{- if .Values.certgen.rbac.annotations }}
     {{- toYaml .Values.certgen.rbac.annotations | nindent 4 -}}
@@ -27,6 +33,12 @@ metadata:
   {{- end }}
   annotations:
     "helm.sh/hook": pre-install, pre-upgrade
+    # ArgoCD deletes a hook that has no delete policy before it makes the
+    # hook again, and it does that on each resume of the operation. The
+    # certgen job needs more time than the interval between two resumes,
+    # so the job never started and envoy-gateway stayed at PreSync. Keep
+    # the hook until it succeeds.
+    "helm.sh/hook-delete-policy": hook-succeeded
     "helm.sh/hook-weight": "-1"   # Ensure rbac is created before the certgen job when using ArgoCD.
   {{- if .Values.certgen.rbac.annotations }}
     {{- toYaml .Values.certgen.rbac.annotations | nindent 4 -}}
@@ -53,6 +65,12 @@ metadata:
   {{- end }}
   annotations:
     "helm.sh/hook": pre-install, pre-upgrade
+    # ArgoCD deletes a hook that has no delete policy before it makes the
+    # hook again, and it does that on each resume of the operation. The
+    # certgen job needs more time than the interval between two resumes,
+    # so the job never started and envoy-gateway stayed at PreSync. Keep
+    # the hook until it succeeds.
+    "helm.sh/hook-delete-policy": hook-succeeded
     "helm.sh/hook-weight": "-1"   # Ensure rbac is created before the certgen job when using ArgoCD.
   {{- if .Values.certgen.rbac.annotations }}
     {{- toYaml .Values.certgen.rbac.annotations | nindent 4 -}}
@@ -78,6 +96,12 @@ metadata:
   {{- end }}
   annotations:
     "helm.sh/hook": pre-install, pre-upgrade
+    # ArgoCD deletes a hook that has no delete policy before it makes the
+    # hook again, and it does that on each resume of the operation. The
+    # certgen job needs more time than the interval between two resumes,
+    # so the job never started and envoy-gateway stayed at PreSync. Keep
+    # the hook until it succeeds.
+    "helm.sh/hook-delete-policy": hook-succeeded
     "helm.sh/hook-weight": "-1"   # Ensure rbac is created before the certgen job when using ArgoCD.
   {{- if .Values.certgen.rbac.annotations }}
     {{- toYaml .Values.certgen.rbac.annotations | nindent 4 -}}
@@ -112,6 +136,12 @@ metadata:
   {{- end }}
   annotations:
     "helm.sh/hook": pre-install, pre-upgrade
+    # ArgoCD deletes a hook that has no delete policy before it makes the
+    # hook again, and it does that on each resume of the operation. The
+    # certgen job needs more time than the interval between two resumes,
+    # so the job never started and envoy-gateway stayed at PreSync. Keep
+    # the hook until it succeeds.
+    "helm.sh/hook-delete-policy": hook-succeeded
     "helm.sh/hook-weight": "-1"   # Ensure rbac is created before the certgen job when using ArgoCD.
   {{- if .Values.certgen.rbac.annotations }}
     {{- toYaml .Values.certgen.rbac.annotations | nindent 4 -}}
--- a/sources/envoy-gateway/v1.8.1/templates/certgen.yaml
+++ b/sources/envoy-gateway/v1.8.1/templates/certgen.yaml
@@ -7,6 +7,12 @@ metadata:
   {{- include "eg.labels" . | nindent 4 }}
   annotations:
     "helm.sh/hook": pre-install, pre-upgrade
+    # ArgoCD deletes a hook that has no delete policy before it makes the
+    # hook again, and it does that on each resume of the operation. The
+    # certgen job needs more time than the interval between two resumes,
+    # so the job never started and envoy-gateway stayed at PreSync. Keep
+    # the hook until it succeeds.
+    "helm.sh/hook-delete-policy": hook-succeeded
   {{- if .Values.certgen.job.annotations }}
     {{- toYaml .Values.certgen.job.annotations | nindent 4 -}}
   {{- end }}
--- a/sources/envoy-gateway/v1.8.1/templates/envoy-proxy-topology-injector-webhook.yaml
+++ b/sources/envoy-gateway/v1.8.1/templates/envoy-proxy-topology-injector-webhook.yaml
@@ -20,6 +20,12 @@ metadata:
   name: 'envoy-gateway-topology-injector.{{ include "eg.namespace" . }}'
   annotations:
     "helm.sh/hook": pre-install, pre-upgrade
+    # ArgoCD deletes a hook that has no delete policy before it makes the
+    # hook again, and it does that on each resume of the operation. The
+    # certgen job needs more time than the interval between two resumes,
+    # so the job never started and envoy-gateway stayed at PreSync. Keep
+    # the hook until it succeeds.
+    "helm.sh/hook-delete-policy": hook-succeeded
     "helm.sh/hook-weight": "-1"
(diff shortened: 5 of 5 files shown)


pre:
## Test on a clean one-node cluster

Kaytoo VM, Ubuntu 24.04, no GPU, `CLUSTER_SIZE: small`,
`DOMAIN: 129.213.21.55.nip.io`, `CERT_OPTION: generate`. Installed with bloom
and then cluster-forge from this branch.

### Correction 1, global.domain: confirmed

| | Before | After |
|---|---|---|
| root app parameters | `clusterForge.targetRevision` only | + `global.domain`, `global.clusterSize` |
| root app valueFiles | `values.yaml` | `values.yaml`, `values_small.yaml` |
| envoy-gateway-config `domain` | no values at all | `129.213.21.55.nip.io` |
| applications | 26 | 53 |
| ComparisonError | 39 of 39 | 0 |

The CoreDNS rewrite is correct, and name resolution now goes both ways as
intended:

```
regex .*\.129.213.21.55.nip.io

argocd-repo-server.argocd.svc.cluster.local  -> 10.243.9.71      (the service)
kc.129.213.21.55.nip.io                      -> 10.243.209.223   (the gateway)
```

Before the correction both names gave the gateway address.

### Correction 2, the PreSync hooks: confirmed

Before, `envoy-gateway` was `Running` for 30 minutes at
`waiting for completion of hook`, and `envoy-gateway-system` held no service
account, no job and no pod. After:

```
phase: Succeeded, message: successfully synced (all tasks run)
pod/envoy-gateway-5f775fc774-tl7bs   1/1   Running
```

The Gateway API CRDs are installed, and both gateways are programmed:

```
ai-gateway   envoy-gateway   10.243.109.13   True
https        envoy-gateway   10.0.255.153    True
```

Applications not `Synced+Healthy` went from 39 to 2.

### This branch alone does not give a cluster that fully settles

The run still failed, and the two remaining faults are NOT corrected here.
They are recorded so that nobody reads this pull request as more than it is.

**a. gitea cannot start on the small path.** bloom skips `bootstrap_gitea.yaml`
when `CLUSTER_SIZE` is small (`clusterforge_setup.yaml:81`), and only that task
makes `gitea-admin-credentials`. `values_small.yaml` still enables `gitea` and
`gitea-config`, so the pod stays at
`secret "gitea-admin-credentials" not found`. Measured: 255 restarts in 59
minutes. A test that removed the two applications from `values_small.yaml` left
one application not settled, so this is the larger of the two. That change is
not in this pull request, because it decides what a small cluster contains.

**b. the AI gateway ext-auth port does not exist.**
`security-policy-ai-gateway-default-deny.yaml:51` asks for port 8083 on
`ai-gateway-system/ai-gateway-discovery`, and envoy-gateway rejects the policy:

```
ExtAuth: TCP Port 8083 not found on service ai-gateway-system/ai-gateway-discovery
```

`ai-gateway-discovery-chart:2.0.0` exposes one port, `health:8081`, and the
container exposes only `health:8081`. There is no ext-auth port at all, so
this is a version mismatch and not a wrong number. Changing 8083 to 8081 would
point ext-auth at the health probe, and `failOpen: false` would then refuse
traffic, so no correction is guessed here.

pre:
## Clean-room test of this branch

One more test, on a machine that never had bloom on it, against the branch as it now stands (`a9615a47`, all three commits).

| Item | Value |
|---|---|
| Machine | fresh Kaytoo VM, Ubuntu 24.04.3, one node, no GPU |
| Start state | no bloom binary, no rke2, no `/etc/bloom` |
| `CLUSTER_SIZE` | `small` |
| Domain | `150.136.144.151.nip.io` |
| bloom | `EAI-8203-ubuntu-2604` |
| cluster-forge | `cf-small-domain-fix-main` |

### The three faults this PR corrects

**1. The root application keeps `global.domain`.** After ArgoCD rendered the chart again to manage the app-of-apps, the live application still holds both value files and all three parameters:

```
valueFiles: ["values.yaml","values_small.yaml"]
parameters: [clusterForge.targetRevision, global.domain, global.clusterSize]
```

CoreDNS got the name it must rewrite, not a wildcard:

```
rewrite continue name regex .*\.150.136.144.151.nip.io https.envoy-gateway-system.svc.cluster.local answer auto
```

Applications in `ComparisonError`: **0**. Before the correction every application was in that state, because the empty domain made CoreDNS rewrite `.*\.` and each name in the cluster went to the gateway.

**2. The envoy-gateway PreSync hooks complete.** The certgen job ran and left its output:

```
envoy-gateway    kubernetes.io/tls   3
envoy            kubernetes.io/tls   3
envoy-rate-limit kubernetes.io/tls   3
```

The application is `Synced` and `Healthy`, and the 10 Gateway API CRDs are installed. Before the correction the job was never created at all: no service account, no job, no pod, for 30 minutes.

**3. gitea is not enabled.** 0 gitea applications and 0 gitea pods. Before, the pod asked for `gitea-admin-credentials` and restarted 255 times in 59 minutes, because bloom skips the task that makes that secret when the size is small.

### Applications

49 of 51 are `Synced` and `Healthy`. The two others are named below and neither belongs to this PR.

### What this test does NOT show

The cluster does not serve traffic yet. Two faults hold it, and this PR does not claim to correct either one.

* **EAI-8231.** `envoy-gateway-config` stays `OutOfSync`. Its one differing resource is `SecurityPolicy/ai-gateway-default-deny`, which asks for port 8083 while the chart gives only `health:8081`.
* **A webhook certificate that does not match its own CA.** Both Gateways stay `Programmed=False` with `NoResources: Envoy replicas unavailable`, because every Envoy proxy pod is refused:

  ```
  failed calling webhook "ai-gateway-controller.envoy-ai-gateway-system.svc.cluster.local":
  x509: certificate signed by unknown authority ... "ai-gateway-controller-ca"
  ```

  The `caBundle` in `envoy-ai-gateway-gateway-pod-mutator` and the `ca.crt` in `self-signed-cert-for-mutating-webhook` are different certificates. The webhook selects `app.kubernetes.io/managed-by=envoy-gateway` on pod CREATE and its `failurePolicy` is `Fail`, so the two proxy Deployments stay at 0/1. The controller restarted 3 times, which points to a race in how it makes its own certificate.

  This comes from `sources/envoy-ai-gateway/v0.6.0/`, which this PR does not change. The hook commit here touches only `sources/envoy-gateway/v1.8.1/`, and that chart has no Helm hooks at all in the ai-gateway source, so it is not the same fault as number 2 above. It needs its own ticket.

pre:
## Test results on one-node Kaytoo VMs

Each run used a fresh VM, cluster-bloom v2.2.1, `DOMAIN=<public-ip>.nip.io`,
`CERT_OPTION=generate`, `GPU_NODE=false`, and a different cluster-forge ref.
bloom finished with `failed=0` in every run. To separate the three fixes, two
temporary refs carried a subset of the commits.

### CLUSTER_SIZE: small

| Ref | Commits | Root app `valueFiles` after it renders itself | gitea apps | envoy-gateway |
|---|---|---|---|---|
| `main` | none of the three | `["values.yaml"]` | 2 | stuck at PreSync |
| `cf-test-no-domain-fix` | hooks + gitea | `["values.yaml"]` | 0 | not reached |
| `cf-test-no-hook-fix` | domain + gitea | `["values.yaml","values_small.yaml"]` | 0 | stuck at PreSync |
| `cf-small-domain-fix-main` | all three | `["values.yaml","values_small.yaml"]` | 0 | Synced / Healthy |

### CLUSTER_SIZE: medium

| Ref | envoy-gateway | Gateway API CRDs |
|---|---|---|
| `main` | Synced / Healthy | 10 |
| `cf-small-domain-fix-main` | Synced / Healthy | 10 |

## What each commit does

### Keep global.domain when the root app renders itself

Necessary. Without this commit the root app drops `values_small.yaml` from its
own spec on the first sync. `externalValues.enabled` then returns to `true`,
and the root app points at the in-cluster Gitea that the small path does not
make. On `main` the medium application set came back, which added the two gitea
applications. On `cf-test-no-domain-fix` the root app went to ComparisonError.

With the commit the two parameters stay in the spec:

```
[{"name":"clusterForge.targetRevision","value":"cf-small-domain-fix-main"},
 {"name":"global.domain","value":"150.136.45.30.nip.io"},
 {"name":"global.clusterSize","value":"values_small.yaml"}]
```

The CoreDNS rewrite then holds the true domain instead of the catch-all
`regex .*\.`:

```
rewrite continue name regex .*\.150.136.45.30.nip.io
        https.envoy-gateway-system.svc.cluster.local answer auto
```

This commit changes only the branch that `externalValues.enabled: false`
selects, and `values_small.yaml` is the one file that sets it false. Medium and
large use the `sources:` branch, where the two values come from
`cluster-values/values.yaml`.

### Keep the envoy-gateway PreSync hooks until they succeed

Necessary for small. `cf-test-no-hook-fix` isolates it, because the domain fix
is present there and the root app kept `values_small.yaml`. The certgen Job
exists, but its ServiceAccount is a hook with weight -1 that ArgoCD removes on
each resume, so the Job cannot make a pod:

```
envoy-gateway-gateway-helm-certgen   Running   0/1   4m35s

Warning FailedCreate (x10 over 4m35s) job-controller
  Error creating: pods "envoy-gateway-gateway-helm-certgen-" is forbidden:
  error looking up service account
  envoy-gateway-system/envoy-gateway-gateway-helm-certgen:
  serviceaccount "envoy-gateway-gateway-helm-certgen" not found
```

At that point the namespace held only the `default` ServiceAccount and no
Gateway API CRD was installed, so each application that uses a Gateway, an
HTTPRoute or a TLSRoute could not sync. With the commit the job completes,
hook-succeeded removes it, and all 10 Gateway API CRDs install.

Note on the scope. Each cluster size uses this chart, but the failure is a race
between the interval of two resumes and the time the certgen job needs. The two
medium runs above show that medium wins that race with the commit and also
without it. So the commit is correct for each size, and the failure appears only
on the small path.

### Stop enabling gitea on the small path

Necessary. On `main` the small run held the `gitea` and `gitea-config`
applications, which wait for a `gitea-admin-credentials` secret that bloom does
not make when `CLUSTER_SIZE` is small. With the branch no gitea application
exists.

## End state of the full branch on small

46 of 49 applications reached Synced / Healthy and stayed there for 13 minutes.
The three others are `amd-gpu-operator`, `amd-gpu-operator-config` and
`kaiwo-config`, which cannot start on a Kaytoo VM because it has no GPU.

brownzebra (approved):
LGTM

PR #794, opened 2026-08-03
Bump envoy + envoy-ai-gateway; cut ai-gateway over to native API-key auth

johnl-amd:
## Summary

Upgrades the gateway stack and moves ai-gateway to native, per-model API-key auth. Bundles the version bump, the auth cutover, the body-path auth supplement, a per-API-key metrics fix, and cleanup of superseded experiments.

**What changed**
- **Version bump:** envoy-gateway `v1.7.1 → v1.8.1`, envoy-ai-gateway `v0.6.0 → v1.0.0`. Old vendored charts removed; `sbom/components.yaml` updated to the new versions.
- **Native API-key auth cutover (EAI-7304):** ai-gateway dropped from the cluster-auth `ext_authz` and enforced by a gateway-scoped default-deny `SecurityPolicy` + per-model `apiKeyAuth` policies (from ai-gateway-discovery). Closes the previous header-less/no-key bypass (verified: no/bogus key with no `x-ai-eg-*` headers → 401).
- **Body-aware `ext_authz` (was #793):** the default-deny SP forwards the request body + `Authorization` to ai-gateway-discovery, which authorizes per-model on the header-less/body path. Adds a cross-namespace `ReferenceGrant`; the discovery namespace is parameterized (`aiGateway.discoveryNamespace`) so the backendRef never drifts.
- **Per-API-key metrics fix:** the v1.0.0 chart shipped `requestHeaderAttributes` / `metricsRequestHeaderAttributes` empty (v0.6.0 populated them), which dropped the controller flags and blanked `api_key_id` / `aim_service_id` on the `gen_ai_*` metrics — breaking AIWB per-key usage stats. Restored in the chart default; `aim_service_id` sourced from `x-ai-eg-backend`.
- **Cleanup:** removed the superseded `set-model-header-from-body` Lua EnvoyExtensionPolicy and its filterOrder wiring, plus stale comments.

**Why**
Keep the gateway stack current, close the API-key auth bypass on the body/header-less path, and restore per-API-key usage metrics in the AIWB UI.

**Root causes (fixes)**
- *Metrics empty:* v1.0.0 chart's empty header-attribute defaults → controller launched without `--metricsRequestHeaderAttributes` → labels never set.
- *Auth bypass:* cluster-auth `ext_authz` did not reject the header-less/no-key path; native apiKeyAuth + default-deny + body-aware ext_authz do.

**Non-obvious decisions**
[...]

PR #817, opened 2026-08-24
Declare the defaulted backendRef weight on ai-gateway-default-deny

johnl-amd:
## Summary

The ext_authz `backendRef` on the `ai-gateway-default-deny` SecurityPolicy does not declare `weight`, but Envoy Gateway defaults it to `1`. ArgoCD owns `spec.extAuth.http` under server-side apply, so the defaulted field lands inside its own field ownership and shows up as a diff it can never reconcile. The SecurityPolicy, and with it the whole `envoy-gateway-config` Application, reports OutOfSync permanently while staying Healthy.

Declaring `weight: 1` makes the manifest match what Envoy Gateway already applies.

**Why:** reported by Petrus as consistently reproducible on freshly deployed clusters from main.

**No behaviour change.** `weight: 1` is the value already in effect. The older `cluster-auth-extauth-policy` has declared `weight` on its backendRef since it was written; this newer policy, added in #794, did not.

## Verification (app-dev, `targetRevision: main`, Envoy Gateway v1.8.1)

It is the only OutOfSync resource in the Application:

```
app: OutOfSync / Healthy
  OUT  SecurityPolicy  ai-gateway-default-deny      OutOfSync
       SecurityPolicy  cluster-auth-extauth-policy  Synced
       ReferenceGrant  ai-gateway-body-authz-grant  Synced
       ...all other 15 resources Synced
```

Field ownership confirms the mechanism:

```
managedFields: manager=argocd-controller op=Apply
               spec.extAuth subfields: [bodyToExtAuth, failOpen, headersToExtAuth, http]
```

`http` is ArgoCD-owned, so the defaulted `weight` beneath it counts against its diff. On `cluster-auth-extauth-policy` ArgoCD owns only `extAuth.grpc`, which is why that policy's own defaulted `failOpen` is ignored and it stays Synced. That asymmetry is why this fix is scoped to one policy.

A server-side dry-run of the fixed manifest, applied as `argocd-controller` against app-dev, produces the live object exactly:

```
kubectl apply --server-side --dry-run=server --field-manager=argocd-controller
diff <dry-run spec> <live spec>  ->  no differences
```

So the diff clears once this lands.

## Scope

Does **not** address the separate ai-gateway-discovery port issue from the same discussion. This policy references the discovery ext_authz handler on port 8083, which first exists in discovery chart 2.0.2; the published chart is 2.0.1 and cluster-forge pins 2.0.0. Header-less inference stays broken until 2.0.2 is cut, promoted, and pinned. Expect the Application to go green while that is still outstanding.

Also unconfirmed: whether the OutOfSync resource on the reporter's fresh clusters is this same one. This is a proven bug on main either way, but the link is worth checking with:

```bash
kubectl get application envoy-gateway-config -n argocd \
  -o jsonpath='{range .status.resources[?(@.status!="Synced")]}{.kind}/{.name}{"\n"}{end}'
```

## Test plan

After merge, `envoy-gateway-config` on app-dev should report Synced with no other resource changing:

```bash
kubectl -n argocd get application envoy-gateway-config -o jsonpath='{.status.sync.status}'
```

### DIFF
--- a/sources/envoy-gateway-config/templates/security-policy-ai-gateway-default-deny.yaml
+++ b/sources/envoy-gateway-config/templates/security-policy-ai-gateway-default-deny.yaml
@@ -49,6 +49,10 @@ spec:
           # backendRef follows wherever discovery deploys — no hardcoded namespace to drift.
           namespace: {{ .Values.aiGateway.discoveryNamespace }}
           port: 8083
+          # Envoy Gateway defaults weight to 1 on every backendRef. Set it explicitly so the
+          # manifest matches the live object, otherwise ArgoCD reports this policy (and the
+          # envoy-gateway-config Application) permanently OutOfSync.
+          weight: 1
       headersToBackend:
         - x-api-key-id
         # Resolved backend (workload id) discovery authorized against — forwarded so the ext_proc


brownzebra (approved):
LGTM

PR #823, opened 2026-08-25
feat(EAI-8292): webhook heal script and platform gates (complements #824)

Q-Dub:
## Summary

Complements #824 (cert-manager webhook TLS prevention) with detect/heal tooling and post-handoff validation. Chart template changes for genCA `caBundle` embedding were dropped — mainline prevention is cert-manager via `root/values.yaml`.

- Add `scripts/ai-gateway-webhook-health.sh` — probe → sync CA from secret → restart controller → roll https + ai-gateway data planes → verify Gateway `Programmed=True`
- Add `scripts/platform-gates/` — Gateway programmed, webhook probe-only, HTTPS AIWB UI, mandatory report artifact
- Wire OpenShift `install.sh` to the shared heal script
- CI: `install-script-checks` job runs offline helper tests (cert-manager webhook render path)

Fixes the k2 incident where `https://aiwbui.<domain>/` failed with connection closed despite healthy app pods — Gateway `Programmed=False`, `Envoy replicas unavailable`.

Jira: https://amd.atlassian.net/browse/EAI-8292

## Test plan

- [x] `scripts/test/run-tests.sh` — cert-manager path renders Certificate, Issuer, and inject-ca-from (no baked caBundle)
- [x] `helm template envoy-ai-gateway sources/envoy-ai-gateway/v1.0.0 --namespace envoy-ai-gateway-system` with mainline `valuesObject` from `root/values.yaml`
- [x] Fresh mainline deploy (k2 / `wookie.silogen.ai`): `gateway/https` and `gateway/ai-gateway` `Programmed=True`
- [x] Webhook probe: `scripts/ai-gateway-webhook-health.sh --probe-only` exits 0
- [x] Envoy data-plane pods: `https` + `ai-gateway` deployments `2/2 Running`
- [x] `https://aiwbui.wookie.silogen.ai/` TLS OK (307 → sign-in) via gateway LB `--resolve`
- [x] `https://argocd.wookie.silogen.ai/` returns 200 via gateway LB `--resolve`
- [x] Drift → heal → data planes recover — validated on chalupa-491a (self-signed drift); **not reproducible on k2** (cert-manager ca-injector + ArgoCD revert caBundle patches immediately)
- [ ] OpenShift install path

### DIFF
--- a/.github/workflows/helm-chart-checks.yaml
+++ b/.github/workflows/helm-chart-checks.yaml
@@ -29,6 +29,20 @@ jobs:
       - name: Helm template
         run: helm template ./root -f ${{ matrix.values }}
 
+  install-script-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Helm
+        uses: azure/setup-helm@v4.3.0
+
+      - name: Run install helper tests
+        run: |
+          chmod +x scripts/ai-gateway-webhook-health.sh scripts/platform-gates/*.sh scripts/test/run-tests.sh
+          scripts/test/run-tests.sh
+
   kyverno-policies:
     runs-on: ubuntu-latest
     strategy:
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ Comprehensive documentation is available in the `/docs` folder:
 | Topic | Documentation |
 |-------|---------------|
 | **Getting Started** | [On-Premises Installation Guide](https://enterprise-ai.docs.amd.com/en/latest/platform-infrastructure/on-premises-installation.html) |
-| **Configuration** | [Cluster Size Configuration](docs/cluster_size_configuration.md) |
+| **Configuration** | [Cluster Size Configuration](docs/cluster_size_configuration.md) · [Configuration Reference](docs/configuration-reference.md) |
 | **Architecture** | [Values Inheritance Pattern](docs/values_inheritance_pattern.md) |
 | **Policy System** | [Kyverno Modular Design](docs/kyverno_modular_design.md) |
 | **Storage Policies** | [Kyverno Access Mode Policy](docs/kyverno_access_mode_policy.md) |
@@ -169,6 +169,8 @@ Comprehensive documentation is available in the `/docs` folder:
 | **CI/CD** | [Workflow Documentation](.github/workflows/README.md) |
 
 Additional documentation:
+- **Configuration reference**: [docs/configuration-reference.md](docs/configuration-reference.md) — Root helmParameters, script env vars, platform gates
+- **Install helpers**: [scripts/README.md](scripts/README.md)
 - **SBOM**: See `/sbom` folder for software bill of materials generation and validation
 
 ## 📝 License
--- a/docs/configuration-reference.md
+++ b/docs/configuration-reference.md
@@ -1,0 +1,60 @@
+# Configuration reference
+
+Cluster-Forge configuration variables used outside Helm chart `values.yaml` files:
+root Argo CD `helmParameters`, install scripts, and platform gate helpers.
+
+## Root chart Argo CD helmParameters
+
+| Application | Parameter | Default (root) | Purpose |
+|-------------|-----------|----------------|---------|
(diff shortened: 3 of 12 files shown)


silokimmo:
Tested this on `chalupa-491a` (RKE2 v1.34.1+rke2r1, single node, cluster size `medium`), which had drifted into exactly the EAI-8292 state. The diagnosis and the heal both check out — one gap to flag.

### Confirmed the root cause matches

Two self-signed CAs generated 4 minutes apart:

| | CN | notBefore |
|---|---|---|
| secret `self-signed-cert-for-mutating-webhook` `ca.crt` | ai-gateway-controller-ca | 10:22:31 |
| webhook `clientConfig.caBundle` | ai-gateway-controller-ca | 10:18:17 |

```
ReplicaSet FailedCreate: failed calling webhook
"ai-gateway-controller.envoy-ai-gateway-system.svc.cluster.local":
x509: certificate signed by unknown authority ... "ai-gateway-controller-ca"
```

`failurePolicy: Fail` meant no pod could be created in `envoy-gateway-system`, so both Envoy data planes were at `0/1` and both Gateways were `Programmed=False reason=NoResources`. Every HTTPS endpoint returned connection refused.

### `ai-gateway-webhook-health.sh` worked

`--probe-only` correctly detected the failure (exit 1). The full heal patched the `caBundle`, restarted the controller, restarted the `https` data plane, re-probed clean, exit 0. `https` went `Programmed=True` and all endpoints came back (argocd/gitea 200, kc 302, openbao/seaweed-admin/airmui/aiwbui 307).

### Gap: `restart_envoy_https_data_plane()` only selects the `https` gateway

`ai-gateway` was broken by the same webhook, but the selector is `gateway.envoyproxy.io/owning-gateway-name=https`, so it was never restarted. Its ReplicaSet had already exhausted its backoff, so it did not self-recover. I polled for three minutes after the heal:

```
t+15s  available=0
t+30s  available=0
...
t+180s available=0
```

Still `0/1`, `Programmed=False`. An explicit `rollout restart` with `owning-gateway-name=ai-gateway` brought it up in ~15s.

The problem is that the script **exits 0 in this state**. The final probe only tests whether the webhook admits pods — which is true — so a cluster can come back half-healed while the script reports success. Same applies to the `gateway-programmed.sh` platform gate if it only checks `https`.

Suggested fix:

```bash
restart_envoy_data_plane() {
  kubectl rollout restart deployment \
    -l 'gateway.envoyproxy.io/owning-gateway-name in (https,ai-gateway)' \
    -n "${GATEWAY_NAMESPACE}" >/dev/null
  kubectl rollout status deployment \
    -l 'gateway.envoyproxy.io/owning-gateway-name in (https,ai-gateway)' \
    -n "${GATEWAY_NAMESPACE}" --timeout=180s
}
```

Or select on label existence (`-l gateway.envoyproxy.io/owning-gateway-name`) to cover any additional Gateways. If the flag is meant to stay https-only, worth renaming it and documenting that other Gateways need separate handling.

### Minor

- The PR description references `scripts/utils/ai-gateway-webhook-health.sh` (summary + last test-plan checkbox), but the file is added at `scripts/ai-gateway-webhook-health.sh`. The code is consistent — `docs/openshift/install.sh` uses the correct path — so it's just stale description text.
- `heal_ai_gateway_webhook()` always restarts `ai-gateway-controller`. On chalupa the controller was already serving the cert matching the secret (verified via `openssl s_client` against `:9443`), so only the `caBundle` patch was needed and the restart cost ~180s. Harmless, and correct in the general case where the controller holds a stale keypair — just noting the patch alone was sufficient here.
- The `caBundle` patch persisted across ArgoCD sync loops, as expected: the current chart doesn't render that field, so there's no desired value to revert to.

Checked off from the test plan against a real drifted cluster:
- [x] `scripts/ai-gateway-webhook-health.sh` exits 0 on healthy cluster
- [x] `kubectl get gateway https -n envoy-gateway-system` shows `Programmed=True`
- [x] Envoy data-plane pods exist
- [x] `https://aiwbui.<domain>/` completes TLS (307)
- [x] `https://argocd.<domain>/` returns 200
- [x] Drift → heal → data plane recovers (with the `ai-gateway` caveat above)

lauri-amd (approved):

PR #829, opened 2026-08-28
fix(blueprint): vllm-sr remove cluster-auth references

lauri-amd:
Cluster-auth has been removed from the stack.
Replaced with envoy SecurityPolicy in the blueprint.


Q-Dub (approved):
LGTM!

PR #831, opened 2026-09-01
docs: align AIM catalog lifecycle with in-tree chart templates

Q-Dub:
Match hardwareFamilies empty vs profiles behaviour to legacy.yaml and profiles.yaml on main; drop the claim that empty list is unsupported.


blankdots (approved):
lgtm

PR #832, opened 2026-09-02
Raise ai-gateway body-authz request-body cap to 32MiB (EAI-8489)

fredespi:
## Problem

The shared Envoy AI Gateway (`ai-gateway`) authorizes header-less requests — model in the JSON body, no `x-ai-eg-model` / `x-ai-eg-backend` routing header — via a body-aware extAuth call to `ai-gateway-discovery`. That extAuth buffers the request body at `maxRequestBytes: 65536` (64 KiB). Per the Envoy Gateway API, a body over that limit returns **HTTP 413 and skips authorization entirely** (this precedence holds even over `failOpen`).

So any large request on the header-less path is rejected before it reaches the model. This is **gateway-wide** — it hits every model on `ai-gateway` (verified on Qwen3.8-27B and MiniMax-M2.5), not one model. Standard OpenAI/Anthropic clients (e.g. Claude Code) put the model in the body and routinely send 200 KB–2 MB requests (a ~200k-token coding turn is ~1 MB; image-bearing turns are larger), so they 413 consistently.

Evidence (direct to the gateway, bypassing the translation proxy): ~67 KB body passes, 120 KB+ returns HTTP 413 (HTTP/2, `text/plain`, Envoy listener shape — not the gateway JSON error). The live Envoy config on the `ai-gateway` pod shows `with_request_body.max_request_bytes: 65536` on the BUFFERED body filter.

## Change

Parameterize the limit as `aiGateway.bodyAuthMaxRequestBytes` (default **32 MiB = 33554432**) and reference it from the `ai-gateway-default-deny` SecurityPolicy. 32 MiB matches clients' own request ceiling and sits under the existing `ai-gateway` `ClientTrafficPolicy` connection `bufferLimit` (50 Mi). The `bodyToExtAuth` API exposes only `maxRequestBytes` (no `allowPartialMessage`), so raising the ceiling is the only lever in the current Envoy Gateway version.

Rendered via `int64` to avoid Helm emitting the value in scientific notation (which would fail the integer schema).

**Tradeoff:** a larger buffer increases the memory the extAuth path can hold per in-flight header-less request. The value is now tunable per environment, so an overlay can lower it.

## Test plan

- [x] `helm template` renders `maxRequestBytes: 33554432` (integer, not `3.3e+07`) when `aiGateway.enabled=true`; renders nothing when disabled.
- [ ] After ArgoCD sync on a cluster with `ai-gateway` enabled: a >64 KB header-less request that previously 413'd is authorized and reaches the model; small-body auth behavior unchanged.

Fixes EAI-8489.

### DIFF
--- a/root/values.yaml
+++ b/root/values.yaml
@@ -12,6 +12,12 @@ externalValues:
 global:
   clusterSize:    # injected via scripts/bootstrap.sh
   domain:         # injected via scripts/bootstrap.sh
+  aiGateway:
+    # Body-authz request-body ceiling, shared by envoy-gateway-config (gateway-scoped policy) and
+    # ai-gateway-discovery (per-model catch-all policies). The catch-all policies are rule-scoped
+    # and override the gateway one per route, so both must carry the same value. Rendered through
+    # int64 below: a bare number here is a float64 and would otherwise emit 4.194304e+06.
+    bodyAuthMaxRequestBytes: 4194304
 localHelmValues:
   enabled: false
 ociRegistry:
@@ -31,6 +37,9 @@ apps:
         value: "ai.{{ .Values.global.domain }}"
       - name: controller.gateway.name
         value: ai-gateway
+      # Same ceiling envoy-gateway-config renders into the gateway-scoped policy.
+      - name: controller.bodyAuthMaxRequestBytes
+        value: "{{ .Values.global.aiGateway.bodyAuthMaxRequestBytes | int64 }}"
   aim-engine:
     repoURL: "{{ .Values.ociRegistry.dockerHub }}"
     repoVersion: "0.2.5"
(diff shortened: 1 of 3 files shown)


johnl-amd:
Pushed a commit on top here, hope that is ok. Two things came out of digging into this.

The gateway-scoped policy is not the one in the path. ai-gateway-discovery also stamps per-model catch-all SecurityPolicies (sectionName route-not-found) with their own hardcoded 65536, and being rule-scoped they win over the gateway one on every model route. On app-dev all three sit at 65536 today, so this PR on its own would not have lifted the 413. Companion: silogen/core#4520, both need to land.

Dropped 32MiB to 4MiB. The authz pod receiving the buffered body is capped at 256Mi memory, and its handler was truncating bodies at 256KiB anyway (fixed in the core PR), so 32MiB was not reachable. 4MiB covers a 200k-token turn with headroom.

Also: wiring the value in root values rendered it as 4.194304e+06, and int64 casts that to 0, which would have quietly published maxRequestBytes: 0. It renders through int64 now and the template fails on a non-positive value.

Title and description still say 32MiB, left those for you.

brownzebra (approved):
LGTM

PR #834, opened 2026-09-03
[EAI-8489]: Raise ai-gateway body-authz request-body cap to 4MiB

johnl-amd:
Backport of #832 (plus the follow-up commit on that branch) to `release/v2.4.x`. Cherry-picked with `-x`, both commits preserved.

- Parameterizes `bodyToExtAuth.maxRequestBytes` on the `ai-gateway-default-deny` SecurityPolicy as `aiGateway.bodyAuthMaxRequestBytes`, default 4MiB. The upstream 64KiB default returns HTTP 413 and skips authorization entirely (ahead of `failOpen`), so every large request on the header-less path was rejected before reaching a model.
- Drives both charts from `global.aiGateway.bodyAuthMaxRequestBytes` in `root/values.yaml`, so the gateway-scoped policy and the per-model catch-all policies stamped by `ai-gateway-discovery` cannot drift. The catch-all policies are rule-scoped and override the gateway-scoped one per route, so a mismatch is silently decided by whichever is lower.
- Renders through `int64` and fails the template on a non-positive value. A bare number in a values file is a float64, so an unguarded render emits `4.194304e+06`, which `int64` casts to `0` and would publish `maxRequestBytes: 0`.

Needs the matching `ai-gateway-discovery` change, silogen/core#4520, to be effective on any cluster with models deployed.

https://jira.amd.com/browse/EAI-8489

## How to test

Send a header-less inference request larger than 64KiB (model in the JSON body, no `x-ai-eg-model` header) with a valid API key. It returned 413 before and should now be served. Verified on app-dev against `MiniMaxAI/MiniMax-M2.5`: 40KB and 200KB return 200, 5MB returns 413 at the new ceiling, and no key returns 401.

### DIFF
--- a/root/values.yaml
+++ b/root/values.yaml
@@ -12,6 +12,12 @@ externalValues:
 global:
   clusterSize:    # injected via scripts/bootstrap.sh
   domain:         # injected via scripts/bootstrap.sh
+  aiGateway:
+    # Body-authz request-body ceiling, shared by envoy-gateway-config (gateway-scoped policy) and
+    # ai-gateway-discovery (per-model catch-all policies). The catch-all policies are rule-scoped
+    # and override the gateway one per route, so both must carry the same value. Rendered through
+    # int64 below: a bare number here is a float64 and would otherwise emit 4.194304e+06.
+    bodyAuthMaxRequestBytes: 4194304
 localHelmValues:
   enabled: false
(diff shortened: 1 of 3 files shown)


Q-Dub (approved):
LGTM

PR #841, opened 2026-09-16
EAI-8713: move OpenShift to otel-lgtm-stack v1.0.8

silokimmo:
## EAI-8713 — move OpenShift to otel-lgtm-stack v1.0.8

> **Stacked PR.** Base is `EAI-8713-otel-lgtm-tuning`, not `main` — applying the same tuning to OpenShift requires the tuned chart. Merge that one first, or rebase this onto main afterwards.

### What changes

`root/values-openshift.yaml`:

1. **`path: otel-lgtm-stack/v1.0.7` → `v1.0.8`** (line 510). OpenShift was the last consumer of v1.0.7 in the ArgoCD path. v1.0.7 has **no Loki retention**, which is the root cause of the Loki PVC fills that had to be expanded twice on other clusters.

2. **Removed 10 `helmParameters` that pinned resources.** These are `--set` values — the highest precedence in Helm — so they would have silently reverted every resource change in the base PR:

   | Removed | Why it mattered |
   |---|---|
   | `collectors.resources.metrics.limits.memory: 4Gi` | Half what metrics-k8s needs; it has OOMKilled at 8Gi elsewhere |
   | `collectors.resources.metrics.requests` (500m/1Gi) | Request below observed usage → permanently scheduled short |
   | `collectors.resources.logs.limits.cpu: "1"` | CPU limit; ~4% throttling measured elsewhere |
   | `lgtm.resources.*` (3 params) | 1 CPU request, 8Gi limit |
   | `lgtm.storage.extra` | Key no longer exists in v1.0.8 |

   **Kept:** `cluster.name`, `dashboards.enabled`, `kubeStateMetrics.enabled`, `nodeExporter.enabled`, `services.nodeExporter.metrics: "9110"` (OpenShift's way around the 9100 conflict), and the four `lgtm.storage.*` sizes.

### Verification

Rendered with exactly the parameters OpenShift still passes: 4 PVCs (no `loki-storage`), node-exporter on 9110, `metricFilters` active, GOMEMLIMIT on the k8s collector, tuned resources, no CPU limits on any collector.

All three `substitutions` were checked against v1.0.8 content and still match — `external-secrets.io/v1beta1` in `grafana-externalsecret.yaml`, and `kube-amd-gpu` plus the `gpu-operator-metrics-exporter` regex in `collectors-metrics-rest.yaml`. A substitution whose pattern stops matching fails silently, which would have quietly dropped an OpenShift-specific fix.

### Caveat

**Not tested on a live OpenShift cluster** — we have no measurements from that environment. Verification here is by render only. Everything applied is chart-default rather than OpenShift-specific, so the blast radius matches the other clusters, but this is the one environment where render is all the evidence we have.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


brownzebra (approved):
OK per @rodrodsilo

PR #842, opened 2026-09-16
EAI-8713: tune otel-lgtm-stack v1.0.8 — drop unused PVC, filter apiserver metrics, size tiers, GOMEMLIMIT

silokimmo:
## EAI-8713 — otel-lgtm-stack v1.0.8 tuning

Four changes to `sources/otel-lgtm-stack/v1.0.8`, plus the size-tier files.

### 1. Remove the unused `loki-storage-pvc`
Proven dead three independent ways: nothing references the mount, the container command `./loki/loki --config.file=...` resolves the **binary** relative to workdir `/otel-lgtm` rather than the `/loki` mount, and Loki writes to `loki-data-pvc`. Roughly **350 GiB reclaimable fleet-wide**.

### 2. Filter apiserver/etcd metrics at scrape time
`kubernetes-apiservers` is **85% of samples and 82% of series** on a measured 3-control-plane cluster. Adds `metric_relabel_configs` to that job, dropping metrics nothing on the platform queries and thinning the two large latency histograms to the buckets actually used.

RE2 has no negative lookahead, so bucket thinning is a two-step mark-then-drop via a `__tmp_*` label (auto-removed after relabeling). `regexQuoteMeta` escapes the bucket values — without it `+Inf` would parse as a quantifier and silently drop the count every quantile depends on.

Profiles: `standard` (default, ~60% of series), `minimal` (~71%), `full` (no filtering). **`docs/otel_lgtm_stack_tuning.md` documents how to re-enable everything** from a cluster's own `cluster-values.yaml`. Dropped samples are never stored, so re-enabling applies going forward only.

No chart dashboard or alert references `apiserver_*` / `etcd_*`.

### 3. Measured requests/limits on every container, split by tier
Several components previously had no requests at all, and the metrics collector's request sat below its observed usage on all seven clusters, so it was permanently scheduled short.

- Memory request = 25% of limit (small/medium), 75% (large)
- **No CPU limits anywhere** — CPU is compressible; throttling a collector only delays scrapes
- small and medium identical (both assumed single-node); large raises lgtm and metrics-k8s to 16Gi

This also **removes the `collectors.resources` / `lgtm.resources` pins from `root/values.yaml`**. Those applied to every tier and would have silently won over the size files, making the tier split a no-op.

### 4. GOMEMLIMIT on the k8s metrics collector
`otelcol_processor_refused_metric_points` is empty on every cluster measured — the memory_limiter has never engaged, so nothing bounded the collector's heap before the cgroup killed it. This tells the runtime the truth about its limit. The existing 6GiB on lgtm is left as-is (inert there: five Go processes share the container, the largest being Prometheus at ~1.65 GiB RSS).

## Testing
Deployed from this branch on an ephemeral OCI VM as a single-node `CLUSTER_SIZE: medium` cluster via cluster-bloom v2.4.1. Verified against live objects: 4 PVCs (no `loki-storage`), tier resources applied through the Gitea `cluster-values` path, zero CPU limits, GOMEMLIMIT present on metrics-k8s, and `metric_relabel_configs` rendered with correct escaping. Collectors ran clean (7/7, no restarts).

Not covered: the 7d retention cycle and the metrics-collector OOMKill, which took 6+ days to appear in the observed incidents — an 8h ephemeral VM cannot reproduce either.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


woojae-siloai (approved):
LGTM

